PR #18, opened 2023-08-06
Show help command if no argument was provided

### LINKED ISSUE #17: Show help command if no argument was provided

myishay:
Fixes #17

### DIFF
--- a/cmd/kor/root.go
+++ b/cmd/kor/root.go
@@ -2,8 +2,9 @@ package kor
 
 import (
 	"fmt"
-	"github.com/spf13/cobra"
 	"os"
+
+	"github.com/spf13/cobra"
 )
 
 var rootCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


yonahd:
Hey looks good.
Can you just sort imports( I need to add this to the CI)

myishay:
Do you mean you prefer the `"github.com/spf13/cobra"` import be between `"fmt"` and `"os"`?

Actually my auto-sorter puts any imports which are not native go imports after the native ones (which is also easier to read IMO).

yonahd (approved):
LGTM.
Thanks for the contribution!

yonahd:
I'll actually make this change in the project and add it to the CI
Thanks!

cmd/kor/root.go
```diff
@@ -12,7 +13,9 @@ var rootCmd = &cobra.Command{
 	Long: `kor is a CLI to to discover unused Kubernetes resources 
 	kor can currently discover unused configmaps and secrets`,
 	Run: func(cmd *cobra.Command, args []string) {
-
+		if len(args) == 0 {
```
luisdavim, line 16: nit: it would be better to set the `Args` property on the `cobra.Command`, something like 
```
Args: cobra.MinimumNArgs(1),
```
would yield  the same result.

yonahd: @luisdavim Sounds good. Can you open an MR for this?

Issue #16, opened 2023-08-04
Add the option to ignore helm secrets from output?


shay-legit:
Hi! Cool project - came across it from natan's post!

This is less of an issue and more of a discussion, but i'd love to have a way to ignore helm secrets from the output.
As you know - [helm v3 uses secrets as its release storage layer](https://helm.sh/docs/faq/changes_since_helm2/#secrets-as-the-default-storage-driver) and i am not sure i would consider those are orphan or unused. perhaps it seem that way but once you will want to perform some helm-related actions, you may need those.

What do you think about that? I don't mind opening a PR once we agree on a small design. e.g we could make it not specific to helm, something like an --exclude option, with a pattern (regex?) being passed that will be used as a filter applied on the  resource name.

in my case it would be something like 

`go run main.go secret --namespace default --exclude 'sh\.helm\.release\.v1\.([a-zA-Z0-9-]+)\.v(\d+)'`

Shay.

yonahd:
This makes sense. Especially since it manages the retention of these for the history and when you uninstall the chart it cleans them up automatically.

Currently, we have a mechanism that works like this(this is in the configmaps) that ignores known false positives
```
var exceptionconfigmaps = []ExceptionResource{
	{ResourceName: "aws-auth", Namespace: "kube-system"},
	{ResourceName: "kube-root-ca.crt", Namespace: "*"},
}
```

We can do something similar in the secrets to exclude the helm release secrets.

PR #20, opened 2023-08-08
Helm secret exception

yonahd:
Create a mechanism to skip secrets based on regex


yonahd:
Feature to support https://github.com/yonahd/kor/issues/16

luisdavim (approved):
LGMT

luisdavim (approved):

PR #22, opened 2023-08-09
feat: add unused HPA detection

shay-legit:


pkg/kor/hpas.go
```diff
@@ -0,0 +1,151 @@
...
+	for _, hpa := range hpas {
+		if hpa.TargetKind == "Deployment" {
+			if !slices.Contains(deploymentNames, hpa.TargetName) {
+				diff = append(diff, hpa.Name)
+			}
+		} else if hpa.TargetKind == "StatefulSet" {
+			if !slices.Contains(statefulsetNames, hpa.TargetName) {
+				diff = append(diff, hpa.Name)
+			}
+		}
```
luisdavim, line 87: nit: use swtich/case

yonahd:
The HPA cmd should be supported in the `all` command as well.

pkg/kor/hpas.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	hpas := make([]Hpa, 0, len(rawHpas.Items))
+	for _, hpa := range rawHpas.Items {
+		hpas = append(hpas, Hpa{
+			TargetKind: hpa.Spec.ScaleTargetRef.Kind,
+			TargetName: hpa.Spec.ScaleTargetRef.Name,
+			Name:       hpa.Name,
+			Namespace:  hpa.Namespace,
+		})
+	}
```
luisdavim, line 59: is the custom `Hpa` type really needed? why not just iterate though the hpa list we get from the API? This way 2 loops are needed.

pkg/kor/hpas.go
```diff
@@ -0,0 +1,151 @@
...
+		} else if hpa.TargetKind == "StatefulSet" {
+			if !slices.Contains(statefulsetNames, hpa.TargetName) {
+				diff = append(diff, hpa.Name)
+			}
+		}
+	}
+	return diff, nil
+}
+
+func ProcessNamespaceHpas(clientset *kubernetes.Clientset, namespace string) ([]string, error) {
```
luisdavim, line 92: does this method need to be public? I don't see it being used outside of this package.

cmd/kor/hpas.go
```diff
@@ -0,0 +1,26 @@
...
+		}
+	},
+}
+
+func init() {
+	hpaCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "k", "", "Path to kubeconfig file (optional)")
+	hpaCmd.PersistentFlags().StringVarP(&namespace, "namespace", "n", "", "Namespace to run on")
+	hpaCmd.PersistentFlags().StringVar(&outputFormat, "output", "table", "Output format (table or json)")
+	rootCmd.AddCommand(hpaCmd)
+}
```
luisdavim, line 26: these flags seem to be getting set on every command, maybe they should be set at the root as `PersistentFlags` so they are available on every child command.

shay-legit:
> The HPA cmd should be supported in the `all` command as well.

good catch! works nicely.

image

cmd/kor/hpas.go
shay-legit: honestly i followed the code base - didn't wanna draft too much from the other resources.
WDYT @yonahd  ?

yonahd: You can leave it this way now. I'll do the refactor across the code base.

pkg/kor/hpas.go
```diff
@@ -0,0 +1,152 @@
...
+		diff, err := processNamespaceHpas(kubeClient, namespace)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to process namespace %s: %v\n", namespace, err)
+			continue
+		}
+		if len(diff) > 0 {
+			if response[namespace] == nil {
+				response[namespace] = make(map[string][]string)
+			}
+			response[namespace]["Hpas"] = diff
```
yonahd, line 141: Can you make this singular
`response[namespace]["Hpa"] = diff`

pkg/kor/all.go
```diff
@@ -80,6 +80,15 @@ func getUnusedRoles(kubeClient *kubernetes.Clientset, namespace string) Resource
 	return namespaceSADiff
 }
 
+func getUnusedHpas(kubeClient *kubernetes.Clientset, namespace string) ResourceDiff {
+	hpaDiff, err := processNamespaceHpas(kubeClient, namespace)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to get %s namespace %s: %v\n", "hpas", namespace, err)
+	}
+	namespaceSADiff := ResourceDiff{"HPA", hpaDiff}
```
yonahd, line 88: Change to lowercase to match the return from the singular json function
namespaceSADiff := ResourceDiff{"Hpa", hpaDiff}

pkg/kor/hpas.go
shay-legit: not a must but if u think it's better without i will change it, no problem

shay-legit: done

yonahd:
Looking good!
Can you just sort your imports using 
`goimports -w .`

yonahd (approved):
LGTM
Thanks for the contribution!

PR #28, opened 2023-08-15
Support finding unused pvcs

yonahd:


pkg/kor/pvc.go
```diff
@@ -0,0 +1,119 @@
...
+		for _, volume := range pod.Spec.Volumes {
+			if volume.PersistentVolumeClaim != nil {
+				usedPvcs = append(usedPvcs, volume.PersistentVolumeClaim.ClaimName)
+			}
+		}
+	}
+	return usedPvcs, err
+}
+
+func calculatePvcDifference(usedConfigMaps []string, configMapNames []string) []string {
```
luisdavim, line 33: the variable names mention `ConfigMaps` but the function is about PVCs, this might be a copy/paste issue, maybe we can just call them `usedObjectNames` and `allObjectNames`?

pkg/kor/pvc.go
```diff
@@ -0,0 +1,119 @@
...
+	var kubeClient *kubernetes.Clientset
+	var namespaces []string
+
+	kubeClient = GetKubeClient(kubeconfig)
+
+	namespaces = SetNamespaceList(namespace, kubeClient)
+	response := make(map[string]map[string][]string)
+
+	for _, namespace := range namespaces {
+		diff, err := processNamespaceHpas(kubeClient, namespace)
```
luisdavim, line 99: shouldn't this be calling `processNamespacePvcs` instead of `processNamespaceHpas`?

Issue #27, opened 2023-08-15
Support kubernetes ingress


yonahd:
Find Kubernetes Ingresses that are not used. -> Not pointing to any services
The command should be `ing` or `ingress`

Yuni-sa:
@yonahd, I can help with this.
what do you think about iterating over the rules and paths of an ingress and checking if the service (using path.Backend.Service.Name) exists?
makes it easy to find unused ingresses by checking if all of the services do not exist, also if one service exist you can skip the rest of the paths in the ingress.

yonahd:
@Yuni-sa Sounds good! 
If the the backend is a resource it should automatically be marked as used (we need to be safe) 
A non prefixed ingress should be accounted for as well 
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-ingress
spec:
  defaultBackend:
    service:
      name: test
      port:
        number: 80

```

yonahd:
Looking at it this way, it is a pretty large pr.
If you want someone to take this together with you it should be an option

PR #31, opened 2023-08-23
Support finding ingresses

Yuni-sa:
to solve #27
check for unused ingresses by checking if services exist, can check also for services in defaultBackend, if ingress points to a resource mark it as used.

### DIFF
--- a/cmd/kor/ingresses.go
+++ b/cmd/kor/ingresses.go
@@ -1,0 +1,25 @@
+package kor
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/yonahd/kor/pkg/kor"
+)
+
+var ingressCmd = &cobra.Command{
+	Use:     "ingress",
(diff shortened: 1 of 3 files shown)


Yuni-sa:
after some more testing I caught a bug where resources are skipped like with this manifest:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: test-ingress
spec:
  rules:
    - host: example.com
      http:
        paths:
          - path: /app1
            pathType: Prefix
            backend:
              service:
                name: service-that-doesnt-exist
                port:
                  number: 8080
          - path: /resource
            pathType: ImplementationSpecific
            backend:
              resource:
                apiGroup: k8s.example.com
                kind: StorageBucket
                name: assets
```

yonahd (approved):
LGTM.
Thanks for the contribution!

Issue #35, opened 2023-08-29
kor doesn't output in JSON format


doryer:
**Describe the bug**
kor cli doesn't output in JSON format while in table mode it is generating output normally

**To Reproduce**
Running: 
kor all -n default --output json

**Expected behavior**
Output in JSON format

**Screenshots**
<img width="596" alt="Screen Shot 2023-08-29 at 12 41 14" src="https://github.com/yonahd/kor/assets/94246801/3047adb0-c3f3-492c-80d9-0d35ed6cf233">

Mac os M1
Kor version: 0.1.9

Yuni-sa:
hi, it looks like in all of the GetUnusedJson() functions for example `GetUnusedDeploymentsJSON()` , the function returns the correct response but does not print it. @yonahd

yonahd:
Yes. This was developed to return a json to other code using kor.

We should support it as output as well.

Yuni-sa:
maybe it should be added in the command files like so: 
```go
if outputFormat == "json" {
	if jsonResponse, err := kor.GetUnusedDeploymentsJSON(namespace, kubeconfig); err != nil {
		fmt.Println(err) // also add some formatting
	} else {
		fmt.Println(jsonResponse)
	}
} else {
	kor.GetUnusedDeployments(namespace, kubeconfig)
}
```
and in that way it doesn't print anything unnecessary when returning the response to code.

yonahd:
@Yuni-sa That should work.
I see that also a couple of the resources are printing the data e.g. https://github.com/yonahd/kor/blob/main/pkg/kor/hpas.go#L127
These should be removed

Can you open a MR for this?

Yuni-sa:
> @Yuni-sa That should work. I see that also a couple of the resources are printing the data e.g. https://github.com/yonahd/kor/blob/main/pkg/kor/hpas.go#L127 These should be removed
> 
good catch! can you asign me to the issue? @yonahd

PR #37, opened 2023-08-29
support json response as cli output

Yuni-sa:
to solve #35 
@yonahd

### DIFF
--- a/cmd/kor/all.go
+++ b/cmd/kor/all.go
@@ -1,6 +1,8 @@
 package kor
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 	"github.com/yonahd/kor/pkg/kor"
 )
@@ -11,7 +13,11 @@ var allCmd = &cobra.Command{
 	Args:  cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		if outputFormat == "json" {
-			kor.GetUnusedAllJSON(namespace, kubeconfig)
+			if jsonResponse, err := kor.GetUnusedAllJSON(namespace, kubeconfig); err != nil {
+				fmt.Println(err)
+			} else {
+				fmt.Println(jsonResponse)
+			}
 		} else {
 			kor.GetUnusedAll(namespace, kubeconfig)
 		}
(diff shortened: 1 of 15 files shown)


Yuni-sa:
only problem i've noticed when testing is that the json output for hpas is different when no unused hpas are found because of a difference in the JSON functions.

output for no unused hpas is: 
```json
{}
```

all other resources when there none unused return something like: 
```json
{
  "kor": {
    "Deployments": null
  }
}
```

refactoring the json outputs to one standard should probably be discussed in a different issue.

Yuni-sa:
> refactoring the json outputs to one standard should probably be discussed in a different issue.

also some returning null and some returning an empty array:
```json
{
  "default": {
    "ConfigMap": [],
    "Deployment": null,
    "Hpa": null,
    "Ingress": [],
    "Pvc": [],
    "Role": [],
    "Secret": [],
    "Service": null,
    "ServiceAccount": [],
    "Statefulset": null
  }
}
```

Yuni-sa:
I can also add `fmt.Fprintf(os.Stderr, "Failed to process namespace %s: %v\n", namespace, err)` to the cmd functions for better formatting.
@yonahd update me on what you think.

yonahd:
> only problem i've noticed when testing is that the json output for hpas is different when no unused hpas are found because of a difference in the JSON functions.
> 
> output for no unused hpas is:
> 
> ```json
> {}
> ```
> 
> all other resources when there none unused return something like:
> 
> ```json
> {
>   "kor": {
>     "Deployments": null
>   }
> }
> ```
> 
> refactoring the json outputs to one standard should probably be discussed in a different issue.

Hmm this is kind of a bug. 
This should definitely be aligned

yonahd:
I'll open an issue to align the json responses

yonahd:
> I can also add `fmt.Fprintf(os.Stderr, "Failed to process namespace %s: %v\n", namespace, err)` to the cmd functions for better formatting. @yonahd update me on what you think.

You mean the tabular response command?

yonahd:
> @Yuni-sa would you like to add this to the MR or should I leave it as a separate issue? #44

Keeping as separate issue

Yuni-sa:
> I'll open an issue to align the json responses

as this got closed by #39 

> @Yuni-sa would you like to add this to the MR or should I leave it as a separate issue? #44

and this is being handled by #44

@yonahd anything more I should fix/add for this pr to be reviewed and merged?

Issue #45, opened 2023-08-31
KUBECONFIG environment variable should be respected


AndiDog:
**Describe the bug**

All "standard" kubectl-like clients respect the KUBECONFIG environment variable, but `kor` does not use it, defaulting to `~/.kube/config`.

**To Reproduce**

```
$ rm ~/.kube/config  # :trollface: - it's my default since I think having multiple dev+prod contexts in one central file is dangerous

$ export KUBECONFIG="${HOME}/.kube/config-some-other-file"

$ kor secret -n kube-system
Failed to load kubeconfig: stat /Users/myname/.kube/config: no such file or directory
```

**Expected behavior**

Respect the environment variable and don't explicitly require `--kubeconfig "$KUBECONFIG"`. If `--kubeconfig` is set though, that argument should have higher priority (as with other clients).

**OS version, architecture and kor version**

macOS Ventura 13.5, arm64
Kor version: 0.1.8

Yuni-sa:
this could be easily solved by changing  `GetKubeClient()` to:

```
if kubeconfig == "" {
	kubeconfig = GetKubeConfigPath()
	if configEnv := os.Getenv("KUBECONFIG"); configEnv != "" {
		kubeconfig = configEnv
	}
}
```
on https://github.com/yonahd/kor/blob/0cf919888a747fd149c99736e6b747245654d8d3/pkg/kor/kor.go#L42

making it default to the  environment variable `KUBECONFIG` and if its empty to use `GetKubeConfigPath()` and still giving `--kubeconfig` his first priority.

yonahd:
> @yonahd might be also good to inclulde the path when recieving an error since now it can come from the environment variable.
> 
> for example changing https://github.com/yonahd/kor/blob/1e0a0a225387425a924537ff00419c701e6babe5/pkg/kor/kor.go#L47C11-L47C11
> 
> to
> 
> ```
> if err != nil {
> 	fmt.Fprintf(os.Stderr, "Failed to load kubeconfig from path %q: %v\n", kubeconfig, err)
> 	os.Exit(1)
> }
> ```

@Yuni-sa No need. The the error can be printed(this is the behavior of kubectl)

rameshgkwd05:
Hello @yonahd I have created a PR please review and let me know if all is well.

This is my first time contributing to your repo. Unfortunately, I could not find any repo specific PR etiquettes.

PR #47, opened 2023-08-31
[Issue 44]  Support yaml format output main feature & issue-45 secondary 

loaynaser3:
also added a fix to the json print issue 
' kor doesn't output in JSON format #35 '


loaynaser3:
@yonahd do you prefer to remove the fixes for issue #35 and #45 ?

yonahd:
> @yonahd do you prefer to remove the fixes for issue #35 and #45 ?

Yes. Please rebase and open a separate MR for #45

cmd/kor/all.go
```diff
@@ -11,7 +13,11 @@ var allCmd = &cobra.Command{
 	Args:  cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		if outputFormat == "json" {
-			kor.GetUnusedAllJSON(namespace, kubeconfig)
+			res, _ := kor.GetUnusedAllJSON(namespace, kubeconfig)
+			fmt.Println(res)
+		} else if outputFormat == "yaml" {
+			res, _ := kor.GetUnusedAllYAML(namespace, kubeconfig)
```
luisdavim, line 19: I don't think a separate function is needed, we can rename `GetUnusedAllJSON` to something like `GetUnusedAllStructured` (or something else) and pass the desired output format.

PR #48, opened 2023-08-31
report if no unused resources found in a namespace

ozgurcd:
Necessary check to detect if there's any unused resource in a namespace. 
Reports a human readable error message if there's not any.

This effects only table output format, json format works as is.


yonahd (approved):
LGTM!
Thanks for the contribution

PR #49, opened 2023-09-02
Read Kubeconfig path from env on priority.

### LINKED ISSUE #45: KUBECONFIG environment variable should be respected

rameshgkwd05:
This PR aims to read Kubeconfig path from environment first and as fallback uses default location.
Methods having only local use are kept/made local.
The unite tests are added to cover newly added code.
fixes #45

### DIFF
--- a/pkg/kor/kor.go
+++ b/pkg/kor/kor.go
@@ -48,7 +48,6 @@ func GetKubeClient(kubeconfig string) *kubernetes.Clientset {
 			kubeconfig = GetKubeConfigPath()
 		}
 	}
-
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to load kubeconfig: %v\n", err)
--- a/pkg/kor/kor.go
+++ b/pkg/kor/kor.go
@@ -33,16 +33,26 @@ func RemoveDuplicatesAndSort(slice []string) []string {
 	return uniqueSlice
 }
 
-func GetKubeConfigPath() string {
+func getDefaultKubeConfigPath() string {
 	home := homedir.HomeDir()
 	return filepath.Join(home, ".kube", "config")
 }
 
-func GetKubeClient(kubeconfig string) *kubernetes.Clientset {
+func loadOrGetKubeConfigPath(kubeconfig string) string {
 	if kubeconfig == "" {
-		kubeconfig = GetKubeConfigPath()
+		if configEnv := os.Getenv("KUBECONFIG"); configEnv != "" {
+			kubeconfig = configEnv
+		} else {
+			kubeconfig = getDefaultKubeConfigPath()
+		}
 	}
-	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	return kubeconfig
+}
+
+func GetKubeClient(kubeconfig string) *kubernetes.Clientset {
+	kubeconfigPath := loadOrGetKubeConfigPath(kubeconfig)
+
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to load kubeconfig: %v\n", err)
 		os.Exit(1)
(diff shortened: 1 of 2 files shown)


AndiDog:
The CLI parameter (`rootCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "k", "", "Path to kubeconfig file (optional)")`) must always take precedence, then environment, then the default path.

pkg/kor/kor.go
```diff
@@ -33,16 +33,26 @@ func RemoveDuplicatesAndSort(slice []string) []string {
...
+		} else {
+			kubeconfig = getDefaultKubeConfigPath()
+		}
 	}
-	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	return kubeconfig
+}
+
+func GetKubeClient(kubeconfig string) *kubernetes.Clientset {
+	kubeconfigPath := loadOrGetKubeConfigPath(kubeconfig)
```
yonahd, line 53: No need for the additional function.
The if else can be done in the original function

yonahd:
> The CLI parameter (`rootCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "k", "", "Path to kubeconfig file (optional)")`) must always take precedence, then environment, then the default path.

Yep. This is done by checking if the value of the flag is empty

AndiDog:
> > The CLI parameter (`rootCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "k", "", "Path to kubeconfig file (optional)")`) must always take precedence, then environment, then the default path.
> 
> Yep. This is done by checking if the value of the flag is empty

Thanks. I wasn't exactly sure from the PR title and by taking a very short look. Change looks good!

yonahd:
@rameshgkwd05 Do you intend to complete this MR?
I would like to put it in the upcoming release

rameshgkwd05:
> @rameshgkwd05 Do you intend to complete this MR? I would like to put it in the upcoming release
Yes @yonahd I would address review comments and make it review ready

pkg/kor/kor.go
rameshgkwd05: addressed this

pkg/kor/kor_test.go
```diff
@@ -60,3 +61,62 @@ func TestCalculateResourceDifference(t *testing.T) {
...
+    cluster: foo-cluster
+    namespace: bar
+  name: foo-context
+current-context: foo-context
+kind: Config
+`
+	return fakeContent
+}
+
+func TestGetKubeClient_fromEnvVar(t *testing.T) {
```
yonahd, line 88: Please rename the function to use camelCase

pkg/kor/kor_test.go
```diff
@@ -60,3 +61,62 @@ func TestCalculateResourceDifference(t *testing.T) {
...
+	defer os.Setenv("KUBECONFIG", originalKCEnv)
+	os.Setenv("KUBECONFIG", configFile.Name())
+
+	kcs := GetKubeClient("")
+	if kcs == nil {
+		t.Errorf("Expected valid clientSet")
+	}
+}
+
+func TestGetKubeClient_fromInput(t *testing.T) {
```
yonahd, line 108: Please rename the function to use camelCase

yonahd (commented):
@rameshgkwd05 Looks good.
A couple of small comments

pkg/kor/kor_test.go
rameshgkwd05: done

pkg/kor/kor_test.go
rameshgkwd05: done

yonahd (approved):
LGTM

PR #57, opened 2023-09-06
Enables to get multiple specific unused resources

Yuni-sa:
to solve #2 

rootcmd now calls for `GetUnusedMulti` and `GetUnusedMultiStructured` when it indicates a list of resources,  `multi.go` uses functions like `getUnusedSVCs` from `all.go`.

@yonahd let me know what you think and also if `--exclude-resource` for `all.go` should be in this PR.

### DIFF
--- a/cmd/kor/root.go
+++ b/cmd/kor/root.go
@@ -3,6 +3,7 @@ package kor
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


yonahd:
Thanks for the MR, I'll review it hopefully in the next few days.
Regarding exclude-resource, I feel like it's an unnecessary option.

luisdavim (approved):
LGTM

Issue #64, opened 2023-09-09
Feature: Run kor in docker


yonahd:
Using a multistage Dockerfile create a Docker image that can run kor.
The final image should be based on alpine

Yuni-sa:
Can I be assigned to this issue?

PR #70, opened 2023-09-11
Add dockerfile

Yuni-sa:
This commit adds a Dockerfile and .dockerignore to for containerizing the kor CLI tool. #64 #69 

example for using the image: `docker run --rm -i -v "$KUBECONFIG:/root/.kube/config" kor-image:tag all -n namespace`
or instead: `docker run --rm -i -v "/path/to/kube/config:/root/.kube/config" kor-image:tag all -n namespace`

The image should be pushed to dockerhub and ghcr.io with every release using goreleaser.

I was also thinking about adding to the Dockerfile `ENV KUBECONFIG=/kubeconfig` making it a bit easier to use the volume: `docker run --rm -i -v "/path/to/kube/config:/kubeconfig" kor-image:tag` but for now users can just add `-e KUBECONFIG=/kubeconfig -v "/path/to/kube/config:/kubeconfig"`.

@yonahd let me know what you think.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.github
+*.yml
+kor
+images
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM golang:1.20.2-alpine AS builder
+
+WORKDIR /build
+COPY . .
+ENV CGO_ENABLED 0
+RUN go build .
+
+FROM alpine:3.18
+
+COPY --from=builder /build/kor /kor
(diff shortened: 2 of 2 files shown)


Yuni-sa:
also instructions should be added to readme.md once the image is pushed.

yonahd (approved):
The if the kubeconfig is mounted it can be passed with the -k flag. So no need to set the env variable.

The MR looks good. I'll add the docker to the release once I create the registry and add credentials

Issue #71, opened 2023-09-11
Inconsistent naming of resource types


Yuni-sa:
**Describe the bug**
different styles of naming resource types and commands. 
some resources accept only single or plural form and some only have the short name.

**for example valid subcommands are:**
ingress - single
services - plural
pdb - shortname
**while they have inconnsistant invalid forms:**
ingresses - plural is invalid
service - single is invalid
poddistruptionbudget/s - both single and plural are invalid

Also nothing is printed when an unsupported subcommand is found making it even more confusing.

**To Reproduce**
try to use the kor cli with commands like service vs services.
`kor service` vs `kor services`  - only plural is valid
`kor secrets` vs `kor secret` - only single form is valid

**Expected behavior**
kor should accept all 3 forms: single, plural and shortnames for supported resources based on what kubectl allows and the api endpoint.

kor should also tell you through rootcmd when it doesn't find a supported resource type.
for example: `kor pv` should output something like `resource type "pv" is not supported use 'kor --help' for available subcommands`

**Screenshots**
![image](https://github.com/yonahd/kor/assets/107778824/03d546dc-d2f2-41f7-8ff3-0c396e7ad063)
![image](https://github.com/yonahd/kor/assets/107778824/a4248ed9-7261-4ac4-9e2f-4595a2557e03)

**OS version, architecture and kor version**
any OS and architecture, tested on windows.
Kor version: v0.2.1

**Additional context**
The reason for this discrapency in form is usually because when using kubectl users usually use the single form (`kubectl get deployment`)  while the api names are in plural form (`GET /apis/apps/v1/deployments`) . in kubectl using deploy, deployment and deployments are all valid options.

PR #72, opened 2023-09-11
Fix inconsistent naming of resource types

Yuni-sa:
#71 
creates a standard for naming resource types: plural, single and shortname. also adding output for unknown/unsupported resource types.


yonahd (approved):
LGTM

Issue #74, opened 2023-09-12
Create CI for deploying docker images


yonahd:
The repo is yonahdissen/kor

PR #73, opened 2023-09-12
Add InCluster support as a CronJob

patricktalmeida:
I liked this tool and wanted to fit one of my use cases, which was to send the output to Slack on a weekly basis.

This PR add a support for `kor` to run as a CronJob inside the cluster and send the output to Slack in 2 ways, either by sending raw text with a Slack webhook, or using hte `files.upload` Slack API to send the output as an attached file which supports formatted and bigger outputs than a webhook.

First of all, sorry about the big PR.
Let me know of your thinking on this.

If it satisfies and you want to have it merged, change the DockerHub repository from `patricktoledodea/kor` to one of yours on both `release.yml` and `values.yaml` as I won't be maintaining the image with all updates.

PR #76, opened 2023-09-12
feat: add docker support with workflow to update image

patricktalmeida:
As discussed in #73, this is a separate PR.

Related issue #74

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,9 +25,6 @@ jobs:
         env:
           GITHUB_TOKEN: "${{ secrets.RELEASE_GITHUB_TOKEN }}"
 
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
-
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -23,4 +23,23 @@ jobs:
         run: |
           goreleaser release
         env:
-          GITHUB_TOKEN: "${{ secrets.RELEASE_GITHUB_TOKEN }}"
\ No newline at end of file
+          GITHUB_TOKEN: "${{ secrets.RELEASE_GITHUB_TOKEN }}"
+
+      - name: Set up QEMU
```
yonahd, line 28: Why do we need to add qemu?

Dockerfile
```diff
@@ -1,12 +1,14 @@
-FROM golang:1.20.2-alpine AS builder
+FROM golang:1.21.0-alpine3.18 as builder
```
yonahd, line 1: We are using go 1.20 currently.
In general you can leave the current dockerfile as is

.github/workflows/release.yml
patricktalmeida: Removed

Dockerfile
patricktalmeida: Removed changes from Dockerfile, but changed `kor` binary path so it can be on $PATH.

yonahd: Sure

Issue #75, opened 2023-09-12
Feature: make kor aware of running in cluster 


yonahd:
When kor runs inside a kubernetes cluster it should mount a service account with Rbac instead of using a kubeconfig.

Make a inCluster flag

PR #77, opened 2023-09-12
feat: add send to slack support

patricktalmeida:
As discussed in #73, this is a separate PR to add send to Slack support.


yonahd:
As a general direction here we should pass a client set to the functions instead of a kubeconfig.
That way we can check if we are in cluster before and either create the clientset from the service account or from the kubeconfig

patricktalmeida:
> As a general direction here we should pass a client set to the functions instead of a kubeconfig. That way we can check if we are in cluster before and either create the clientset from the service account or from the kubeconfig

@yonahd I didn't quite understand your idea. I created 2 separate handlers for InCluster and KubeConfig configurations.
I create the clientset from those handlers and I let `GetKubeClient` handle which one to use. And kubeconfig file is only sent to `KubeConfigClient` which needs it to create the client.

Could you explain your idea when looking into those 3 functions? `KubeConfigClient`, `InClusterConfigClient` and `GetKubeClient`

yonahd:
> > As a general direction here we should pass a client set to the functions instead of a kubeconfig. That way we can check if we are in cluster before and either create the clientset from the service account or from the kubeconfig
> 
> @yonahd I didn't quite understand your idea. I created 2 separate handlers for InCluster and KubeConfig configurations. I create the clientset from those handlers and I let `GetKubeClient` handle which one to use. And kubeconfig file is only sent to `KubeConfigClient` which needs it to create the client.
> 
> Could you explain your idea when looking into those 3 functions? `KubeConfigClient`, `InClusterConfigClient` and `GetKubeClient`

I don't want to limit any of the functions to only be in cluster or only out of cluster. That's why I want to create the separation of kubeconfig vs service account earlier and regardless pass a clientset to the functions.

patricktalmeida:
> > > As a general direction here we should pass a client set to the functions instead of a kubeconfig. That way we can check if we are in cluster before and either create the clientset from the service account or from the kubeconfig
> > 
> > 
> > @yonahd I didn't quite understand your idea. I created 2 separate handlers for InCluster and KubeConfig configurations. I create the clientset from those handlers and I let `GetKubeClient` handle which one to use. And kubeconfig file is only sent to `KubeConfigClient` which needs it to create the client.
> > Could you explain your idea when looking into those 3 functions? `KubeConfigClient`, `InClusterConfigClient` and `GetKubeClient`
> 
> I don't want to limit any of the functions to only be in cluster or only out of cluster. That's why I want to create the separation of kubeconfig vs service account earlier and regardless pass a clientset to the functions.

I think I got your idea. But still thinking on how to control that. I mean, when it should authenticate with kubeconfig or when to use `rest.InClusterConfig()`.
What do you think of sending a flag when it should authenticate from inside the cluster, e.g. `kor all --in-cluster-auth`?
With that it could be could be hard coded inside the helm template for CronJob and also in the future for Deployments. e.g. `args: ["{{ .Values.cronJob.korCommand }} --in-cluster-auth --slack-channel {{ .Values.cronJob.slackChannel }} --slack-auth-token {{ .Values.cronJob.slackAuthToken }}"]`

yonahd:
We can try something like this 
```
func GetKubeClient(kubeconfig string) *kubernetes.Clientset {
	if _, err := os.Stat("/var/run/secrets/kubernetes.io/serviceaccount/token"); err == nil {
		config, err := rest.InClusterConfig()
		if err != nil {
			fmt.Fprintf(os.Stderr, "Failed to load kubeconfig: %v\n", err)
			os.Exit(1)
		}

		clientset, err := kubernetes.NewForConfig(config)
		if err != nil {
			fmt.Fprintf(os.Stderr, "Failed to create Kubernetes client: %v\n", err)
			os.Exit(1)
		}
		return clientset
	}
	if kubeconfig == "" {
		if configEnv := os.Getenv("KUBECONFIG"); configEnv != "" {
			kubeconfig = configEnv
		} else {
			kubeconfig = GetKubeConfigPath()
		}
	}
	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
	if err != nil {
		fmt.Fprintf(os.Stderr, "Failed to load kubeconfig: %v\n", err)
		os.Exit(1)
	}

	clientset, err := kubernetes.NewForConfig(config)
	if err != nil {
		fmt.Fprintf(os.Stderr, "Failed to create Kubernetes client: %v\n", err)
		os.Exit(1)
	}
	return clientset
}
```

and in the command we can do e.g.
```
var configmapCmd = &cobra.Command{
	Use:     "configmap",
	Aliases: []string{"cm"},
	Short:   "Gets unused configmaps",
	Args:    cobra.ExactArgs(0),
		clientset := kor.GetKubeClient(kubeconfig)
		if outputFormat == "json" || outputFormat == "yaml" {
			if response, err := kor.GetUnusedConfigmapsStructured(includeExcludeLists, clientset, outputFormat); err != nil {
				fmt.Println(err)
			} else {
				fmt.Println(response)
			}
		} else {
			kor.GetUnusedConfigmaps(includeExcludeLists, clientset)
		}

	},
}

```

patricktalmeida:
> We can try something like this
> 
> ```
> func GetKubeClient(kubeconfig string) *kubernetes.Clientset {
> 	if _, err := os.Stat("/var/run/secrets/kubernetes.io/serviceaccount/token"); err == nil {
> 		config, err := rest.InClusterConfig()
> 		if err != nil {
> 			fmt.Fprintf(os.Stderr, "Failed to load kubeconfig: %v\n", err)
> 			os.Exit(1)
> 		}
> 
> 		clientset, err := kubernetes.NewForConfig(config)
> 		if err != nil {
> 			fmt.Fprintf(os.Stderr, "Failed to create Kubernetes client: %v\n", err)
> 			os.Exit(1)
> 		}
> 		return clientset
> 	}
> 	if kubeconfig == "" {
> 		if configEnv := os.Getenv("KUBECONFIG"); configEnv != "" {
> 			kubeconfig = configEnv
> 		} else {
> 			kubeconfig = GetKubeConfigPath()
> 		}
> 	}
> 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
> 	if err != nil {
> 		fmt.Fprintf(os.Stderr, "Failed to load kubeconfig: %v\n", err)
> 		os.Exit(1)
> 	}
> 
> 	clientset, err := kubernetes.NewForConfig(config)
> 	if err != nil {
> 		fmt.Fprintf(os.Stderr, "Failed to create Kubernetes client: %v\n", err)
> 		os.Exit(1)
> 	}
> 	return clientset
> }
> ```
> 
> and in the command we can do e.g.
> 
> ```
> var configmapCmd = &cobra.Command{
> 	Use:     "configmap",
> 	Aliases: []string{"cm"},
> 	Short:   "Gets unused configmaps",
> 	Args:    cobra.ExactArgs(0),
> 	Run: func(cmd *cobra.Command, args []string) {
> 		clientset := kor.GetKubeClient(kubeconfig)
> 		if outputFormat == "json" || outputFormat == "yaml" {
> 			if response, err := kor.GetUnusedConfigmapsStructured(includeExcludeLists, clientset, outputFormat); err != nil {
> 				fmt.Println(err)
> 			} else {
> 				fmt.Println(response)
> 			}
> 		} else {
> 			kor.GetUnusedConfigmaps(includeExcludeLists, clientset)
> 		}
> 
> 	},
> }
> ```

Agreed and done. Let me know if you need to change something.

yonahd:
Perfect!
Don't want to inconvenience you too much, but can we separate the slack part to a different MR?

patricktalmeida:
> Perfect! Don't want to inconvenience you too much, but can we separate the slack part to a different MR? This MR would close #75 And I'll open a different ticket for Slack messages(I think we can combine some of the functions). In addition We can have the slack reports also out of the cluster

Sure. I can split it in two.
What do you mean by having slack report out of the cluster? I mean it already works out of the cluster also. If you run it from your local machine `kor all --slack-channel <your-channel> --slack-auth-token <your-token> --include-namespaces default,some-ns` it's gonna work just fine.

yonahd:
> > Perfect! Don't want to inconvenience you too much, but can we separate the slack part to a different MR? This MR would close #75 And I'll open a different ticket for Slack messages(I think we can combine some of the functions). In addition We can have the slack reports also out of the cluster
> 
> Sure. I can split it in two. What do you mean by having slack report out of the cluster? I mean it already works out of the cluster also. If you run it from your local machine `kor all --slack-channel <your-channel> --slack-auth-token <your-token> --include-namespaces default,some-ns` it's gonna work just fine.

My mistake.
Now though it can be a condition in the regular function.

yonahd:
I'd like to condense the slack functions to be part of the `GetUnusedHpas` function.
Can we make the slack parameters optional and based on the users sent parameters decide if we are sending to slack(and whether it's a file or webhook)?

patricktalmeida:
> I'd like to condense the slack functions to be part of the `GetUnusedHpas` function. Can we make the slack parameters optional and based on the users sent parameters decide if we are sending to slack(and whether it's a file or webhook)?



> I'd like to condense the slack functions to be part of the `GetUnusedHpas` function. Can we make the slack parameters optional and based on the users sent parameters decide if we are sending to slack(and whether it's a file or webhook)?

alright, give me some time to take a look at that.

pkg/kor/all.go
```diff
@@ -144,9 +148,25 @@ func GetUnusedAll(includeExcludeLists IncludeExcludeLists, clientset kubernetes.
...
+
 		output := FormatOutputAll(namespace, allDiffs)
-		fmt.Println(output)
-		fmt.Println()
+
+		outputBuffer.WriteString(output)
+		outputBuffer.WriteString("\n")
+	}
+
+	if len(slackParams) == 1 {
```
luisdavim, line 158: instead of relying on the length and order of the passed params I think it would be better to pass a single `slackOpts` struct and check what's set or not there, this also avoids the `if`/`else` chain in the cobra `Run` functions when calling these methods.

luisdavim: it should also be possible to consolidate `SendToSlackWebhook` and `SendFileToSlack` into a single `SendToSlack` function and handle everything from there based on what's being set in the `slackOpts` structure.

patricktalmeida: Agreed. Gonna work on these.

yonahd:
@patricktalmeida Is this ready for review?

patricktalmeida:
> @patricktalmeida Is this ready for review?

Yes it is.

pkg/kor/slack.go
```diff
@@ -0,0 +1,110 @@
...
+	return nil
+}
+
+func writeOutputToFile(outputBuffer string) (string, error) {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get user's home directory: %v", err)
+	}
+
+	outputFileName := "output.txt"
```
yonahd, line 95: Can you rename the file to kor-scan-results.txt

pkg/kor/slack.go
```diff
@@ -0,0 +1,110 @@
...
+func (sm SlackMessage) SendToSlack(slackOpts SlackOpts, outputBuffer string) error {
+	if slackOpts.WebhookURL != "" {
+		payload := []byte(`{"text": "` + outputBuffer + `"}`)
+		_, err := http.Post(slackOpts.WebhookURL, "application/json", bytes.NewBuffer(payload))
+
+		if err != nil {
+			return err
+		}
+		return nil
+	} else if slackOpts.Channel != "" && slackOpts.Token != "" {
```
yonahd, line 39: Let's add a print saying sending to channel %s

cmd/kor/root.go
```diff
@@ -39,13 +39,17 @@ var (
...
+	slackOpts           kor.SlackOpts
 )
 
 func Execute() {
 	rootCmd.PersistentFlags().StringVarP(&kubeconfig, "kubeconfig", "k", "", "Path to kubeconfig file (optional)")
 	rootCmd.PersistentFlags().StringVarP(&includeExcludeLists.IncludeListStr, "include-namespaces", "n", "", "Namespaces to run on, splited by comma. Example: --include-namespace ns1,ns2,ns3. ")
 	rootCmd.PersistentFlags().StringVarP(&includeExcludeLists.ExcludeListStr, "exclude-namespaces", "e", "", "Namespaces to be excluded, splited by comma. Example: --exclude-namespace ns1,ns2,ns3. If --include-namespace is set, --exclude-namespaces will be ignored.")
 	rootCmd.PersistentFlags().StringVar(&outputFormat, "output", "table", "Output format (table or json)")
+	rootCmd.PersistentFlags().StringVar(&slackOpts.WebhookURL, "slack-webhook-url", "", "Slack webhook URL to send notifications to")
+	rootCmd.PersistentFlags().StringVar(&slackOpts.Channel, "slack-channel", "", "Slack channel to send notifications to")
```
yonahd, line 51: It should be stated that a channel without a token won't work(the same for the token)

yonahd (commented):
Overall looks good.
If you can add a screenshot from slack to images(otherwise I'll do it)
Also there is the Readme updates which you can optionally do

yonahd:
FYI I also opened this.
https://github.com/yonahd/kor/issues/91
We might have to make a small refactor in the code to support it in Helm properly

patricktalmeida:
> FYI I also opened this. #91 We might have to make a small refactor in the code to support it in Helm properly

I'm gonna look into that.

patricktalmeida:
> Overall looks good. If you can add a screenshot from slack to images(otherwise I'll do it) Also there is the Readme updates which you can optionally do

Well, I believe it would be better if you added it. The cluster I'm testing on has sensitive data.

yonahd (approved):
LGTM

PR #78, opened 2023-09-12
feat: add helm charts with cronjob support

patricktalmeida:


charts/kor/values.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+cronJob:
+  enabled: true
+  name: kor
+  schedule: "0 1 * * 1"
+  image:
+    repository: yonahdissen/kor
+    tag: latest
+  # e.g. kor configmap --include-namespace default,other-ns
+  korCommand: kor all
```
yonahd, line 11: Lets change this to command

charts/kor/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "0.1.0"
```
yonahd, line 24: Can you put in the actual version of the app

yonahd (commented):
Looks good overall
Please generate the readme for the Chart using 
docker run --rm --volume "$(pwd):/helm-docs" -u "$(id -u)" jnorwood/helm-docs:v1.8.1

charts/kor/values.yaml
patricktalmeida: My idea was to have it configurable so if someone doesn't want to run `kor all` and instead run only `kor secrets` for example they could configure it through helm cli. e.g.

```
helm upgrade -i kor \
    --namespace kor \
    --create-namespace \
    --set cronJob.slackChannel=<slack-channel> \
    --set cronJob.slackToken=<slack-token> \
    --set cronJob.korCommand="kor secrets --include-namespace example-ns"
    ./charts/kor
``` 

If you still want it, I can hard code it into the CronJob definition as `kor all`. Let me know what you think.

yonahd: I just mean the variable name

patricktalmeida: Oh, I completely misunderstood it 🤦

charts/kor/values.yaml
```diff
@@ -0,0 +1,26 @@
+namespace: kor
```
yonahd, line 1: Can you remove the namespace from the values? (let the user set it through the -n flag)

charts/kor/templates/cronjob.yaml
```diff
@@ -0,0 +1,37 @@
...
+          serviceAccountName: {{ include "kor.serviceAccountName" . }}
+          containers:
+            - name: {{ .Release.Name }}-container
+              image: {{ .Values.cronJob.image.repository }}:{{ .Values.cronJob.image.tag }}
+              command: ["/bin/sh", "-c"]
+              {{- if .Values.cronJob.slackWebhookUrl }}
+              args: ["{{ .Values.cronJob.command }} --slack-webhook-url {{ .Values.cronJob.slackWebhookUrl }}"]
+              {{- else if and .Values.cronJob.slackChannel .Values.cronJob.slackAuthToken }}
+              args: ["{{ .Values.cronJob.command }} --slack-channel {{ .Values.cronJob.slackChannel }} --slack-auth-token {{ .Values.cronJob.slackAuthToken }}"]
+              {{- end }}
```
yonahd, line 25: Please allow also running without slack notifications
```
{{- else }}
args: ["{{ .Values.cronJob.command }}"]
```

charts/kor/values.yaml
patricktalmeida: values file is just a default value, the user can modify it by passing it through the helm command line on installation/upgrade time.

If I remove this default value and leave only the flag -n on helm command line it would place the helm chart creation on kor namespace but subjects ClusterRoleBinding need the namespace value to be deployed.
I removed it locally and tested and got this error:

```
Error: UPGRADE FAILED: cannot patch "kor-read-resources-clusterrolebinding" with kind ClusterRoleBinding: ClusterRoleBinding.rbac.authorization.k8s.io "kor-read-resources-clusterrolebinding" is invalid: subjects[0].namespace: Required value
```

If I remove it the user would need to also pass the set flag `--set namespace=kor` on helm command line:

```
helm upgrade -i kor --namespace kor --create-namespace --set namespace=kor --set cronJob.slackChannel=<slack-channel> --set cronJob.slackAuthToken=<user-token> ./charts/kor
```

What do you think is best? I think it's redundant to have both `--set namespace=kor` and `--namespace kor` in the same command. Even though the user can still do it with the default value set to `namespace: kor`.

yonahd (approved):
LGTM

PR #87, opened 2023-09-21
Export Prometheus metrics

loaynaser3:
adding option for "exporter"
this is an initial PR, simple way of exporting metrics
@yonahd  , @Yuni-sa  how you think we should proceed from here ?


Yuni-sa:
@loaynaser3 @yonahd 
this looks really good I think I have everything else set up regarding the helm packaging and deployment can I add it here?

yonahd:
> @loaynaser3 @yonahd this looks really good I think I have everything else set up regarding the helm packaging and deployment can I add it here or should I create I different pr for the helm chart?

Let's put that in a separate PR

yonahd:
> also, it's probably best to source the time interval from an environment variable, allowing for easy changes from the helm chart.

+1
Just create some default value if the environment variable doesn't exist.

pkg/kor/exporter.go
```diff
@@ -0,0 +1,73 @@
...
+	http.Handle("/metrics", promhttp.Handler())
+	fmt.Println("Server listening on :8080")
+	go exportMetrics(includeExcludeLists, clientset, outputFormat) // Start exporting metrics in the background
+	if err := http.ListenAndServe(":8080", nil); err != nil {
+		fmt.Println(err)
+	}
+}
+
+func exportMetrics(includeExcludeLists IncludeExcludeLists, clientset kubernetes.Interface, outputFormat string) {
+	exporterInterval := os.Getenv("exporterInterval")
```
Yuni-sa, line 41: Can you convert environment variable names to uppercase and separated by underscores? For example: `EXPORTER_INTERVAL`

cmd/kor/exporter.go
```diff
@@ -0,0 +1,21 @@
+package kor
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/yonahd/kor/pkg/kor"
+)
+
+var exporterCmd = &cobra.Command{
+	Use:   "exporter",
```
patricktalmeida, line 9: Wouldn't it be better if this command specified the destination where it's exporting to? I mean maybe in the future some other feature could try to export somewhere else other then to Prometheus.
What do you think of having something like:

`kor exporter --prometheus`
or
`kor prom-exporter`

loaynaser3: sounds good, but since we only have Prometheus support no need for a change
we can keep it the default unless specified other, same as in the output format .. 
WDYT ?

yonahd:
@loaynaser3 Looking over the implementation now.
There is a use case that comes to mind where a user want to run kor only for some resources and would still like to get metrics. What would be the cost of implementing this as a flag for the commands?
e.g.
`kor cm,secret --metrics`
or something like that

loaynaser3:
@yonahd i thought of this scenario but i didnt see the use of it. 
regarding this change its not a major change, we can add a call function to root and change the if condition to switch case

yonahd:
> @yonahd i thought of this scenario but i didnt see the use of it. regarding this change its not a major change, we can add a call function to root and change the if condition to switch case

Ok. Let's keep it this way for now.
Is this ready for review?

loaynaser3:
@yonahd yes

pkg/kor/exporter.go
```diff
@@ -0,0 +1,73 @@
...
+			var data map[string]map[string][]string
+			if err := json.Unmarshal([]byte(korOutput), &data); err != nil {
+				fmt.Println("Error parsing JSON:", err)
+				return
+			}
+
+			for namespace, resources := range data {
+				for kind, resourceList := range resources {
+					for _, resourceName := range resourceList {
+						// orphanedResourcesCounter.WithLabelValues(kind, namespace, resourceName).Inc()
```
yonahd, line 65: Please remove this comment

pkg/kor/exporter.go
```diff
@@ -0,0 +1,73 @@
...
+	"strconv"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"k8s.io/client-go/kubernetes"
+)
+
+var (
+	orphanedResourcesCounter = prometheus.NewCounterVec(
```
yonahd, line 17: I think the correct type here would be a gauge(since we are not counting), and set the value to 1.
WDYT?

yonahd (commented):
@loaynaser3 Looks good.
A couple small comments

pkg/kor/exporter.go
loaynaser3: yea it should be either gauge or SetToCurrentTime()

pkg/kor/exporter.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"k8s.io/client-go/kubernetes"
+)
+
+var (
+	orphanedResourcesCounter = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "Kubernetes_Orphaned_Resources",
```
yonahd, line 19: Metric names should be lower case

yonahd:
@loaynaser3 Can you also change the name of the pull request

yonahd (approved):
LGTM

PR #88, opened 2023-09-21
Update installation instructions to README.md

Yuni-sa:
Add instructions for installing with `go install` and docker.


yonahd (approved):
LGTM

Issue #40, opened 2023-08-30
feature request: Export Prometheus metrics


Razgoldenberg1:
**Summary**
I would like to suggest that the Kor project be updated to support Prometheus metrics.
CLI is very useful but allows users to collect and monitor Kor's metrics using Prometheus giving consistent visibility to the Kubernetes Orphaned Resources on your cluster.

yonahd:
Hi @Razgoldenberg1 
What kind of metrics do you propose to export?

Razgoldenberg1:
Hi @yonahd thanks for the quick response,
To start I would like to get the same output as on the CLI .
if we could get the name of the resource, namespace, and kind it would be very nice:
something like that:
Kubernetes_Orphaned_Resources{kind="configmap", namespace="default" ,name="my_configmap", age="265days"}

yonahd:
Interesting proposal.
Will look into this.

loaynaser3:
@yonahd  have a look at this [commit](https://github.com/yonahd/kor/commit/611e4add7b63373a9bf47cdece8841a251c3f118 )
may be helpful for this issue |output example:
```
Kubernetes_Orphaned_Resources{kind="Service",namespace="okteto",resourceName="rafiki"} 0
Kubernetes_Orphaned_Resources{kind="Secret",namespace="prometheus",resourceName="grafana-tls-certificate"} 0
```
regarding adding the age of the resource still not available

yonahd:
Thanks!
This feature doesn't make so much sense as this is just a cli tool at this point.
This feature will be useful down the road when we support running this as a server. But for now this is in the backlog.

yonahd:
@loaynaser3 this is now more relevant.
Would you like it assigned to you?

Yuni-sa:
I can also help with this if needed, would be better to start working on it after #78 is merged.

yonahd:
Thanks @Yuni-sa!
I'll create a release(with a docker image) and then we can move forward with the Helm chart.

Regarding the metrics I don't see a reason to complicate with the push gateway. We can have the deployment offer metrics or/and slack messages, While the cron will only offer slack.

loaynaser3:
> @loaynaser3 this is now more relevant. Would you like it assigned to you?

yea sure, you can assign it

yonahd:
> > @loaynaser3 this is now more relevant. Would you like it assigned to you?
> 
> yea sure, you can assign it

Assigned. I beleive @Yuni-sa can assist or take it if you need

PR #90, opened 2023-09-21
Update helm charts with deployment support

Yuni-sa:
for #40 and #87.

### DIFF
--- a/charts/kor/values.yaml
+++ b/charts/kor/values.yaml
@@ -53,6 +53,22 @@ prometheusExporter:
     type: ClusterIP
     port: 8080
 
+  serviceMonitor:
+    enabled: true
+    # Set the namespace the ServiceMonitor should be deployed, if empty namespace will be .Release.Namespace
+    namespace: ""
+    # Service monitor labels
+    labels: {}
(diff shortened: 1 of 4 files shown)


yonahd:
Can you add a ServiceMonitor to the chart

charts/kor/values.yaml
```diff
@@ -14,6 +14,45 @@ cronJob:
...
+prometheusExporter:
+  enabled: true
+  name: kor-exporter
+  # time in minutes, default is 10 minutes
+  exporterInterval: ""
+  deployment:
+    image:
+      repository: yonahdissen/kor
+      tag: latest
+    command: kor exporter
```
patricktalmeida, line 26: Wouldn't it be better if this command specified the destination where it's exporting to? I mean maybe in the future some other feature could try to export somewhere else other then to Prometheus.
What do you think of having something like:

`kor exporter --prometheus`
or
 `kor prom-exporter`

yonahd:
LGTM!
Waiting for https://github.com/yonahd/kor/pull/87 before merging

yonahd (approved):
LGTM

PR #101, opened 2023-10-10
Fix: Add helm condition to skip secret manifest creation when slack secrets are not provided

### LINKED ISSUE #100: Bug: Helm chart creates secrets even when no slack token is given
**Describe the bug**
Helm creates unused Secrets https://github.com/yonahd/kor/blob/main/charts/kor/templates/secret.yaml 
when slack token is not given

**To Reproduce**
Run Helm install without providing slack tokens

e.g. Darwin x86_64
Kor version: 0.2.4

Yuni-sa:
fix #100 
@yonahd


yonahd (approved):
LGTM

PR #102, opened 2023-10-13
Support finding unused namespaces

### LINKED ISSUE #92: Find Empty namespaces
Find namespaces with no resources

liornoy:
Add a new "namespace" command to find unused namespaces.

This command determines whether there are any used resources
within a given namespace, if there are no used resources,
the namespace is considered unused.

fixes #92

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Kor is a tool to discover unused Kubernetes resources. Currently, Kor can identi
 - PVCs
 - Ingresses
 - PDBs
+- Namespaces
 
 ![Kor Screenshot](/images/screenshot.png)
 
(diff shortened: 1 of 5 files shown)


yonahd:
Hi @liornoy
This is a good direction but this would miss some resources that may be present such a jobs, cron jobs or any other resource kor doesn't track.

luisdavim:
> Hi @liornoy This is a good direction but this would miss some resources that may be present such a jobs, cron jobs or any other resource kor doesn't track.

Yeah,  this may lead to many false positives, one approach could be to get a list of resource types from the user and check if any of those exist in the given namespace, or get a list of namespaced resource types from the API and use that to check the NS. Something like what this plugin does https://github.com/corneliusweig/ketall/tree/master

We also have to take into account the default resources in a namespace,  like the default service account and its secret for example.

liornoy:
Thanks for the review @yonahd @luisdavim !
I'll take a look into this `ketall` tool and rethink my approach here.

yonahd:
@liornoy Do you intend to retry this MR or should I have someone else attempt it?

liornoy:
@yonahd Hi! please do un-assign me. I can't seem to make this work in the time that I have.

yonahd:
Thanks for the attempt!
Hope to see you back here in the future

roopeshsn:
Hi, @yonahd! I want to attempt to solve this issue. I have a question. Whether we know exactly what are the default resources including service account, `kube-root-ca.crt` configmap etc. or do those resources change in terms of distributions?

Issue #103, opened 2023-10-15
Feature: Support for Identifying Unused CRDs


m-yosefpor:
**Is your feature request related to a problem? Please describe.**
While Kor is an excellent tool for identifying unused Kubernetes resources, it currently does not support identifying unused Custom Resource Definitions (CRDs). In large Kubernetes clusters, it's not uncommon to have CRDs that were once used but no longer have any instances created from them. Identifying these unused CRDs can help in cleaning up and maintaining a lean cluster configuration.

**Describe the solution you'd like**
I'd like Kor to have the capability to identify and list unused CRDs, i.e., CRDs for which no instances have been created. This can be an additional subcommand, similar to the existing ones like `configmap`, `secret`, etc. For example: `kor crd` could list all the unused CRDs in the specified namespace or all namespaces.

**Describe alternatives you've considered**
One could manually check for CRD instances using `kubectl`, but this is cumbersome and not scalable for large clusters. Having this feature integrated into Kor would streamline the process and make it consistent with the identification of other unused resources.

**Additional context**
As Kubernetes continues to evolve and organizations adopt more custom resources, the ability to identify unused CRDs will become increasingly important for cluster hygiene. Adding this feature would further enhance Kor's value proposition as a comprehensive Kubernetes resource management tool.

yonahd:
Thanks for the proposal! 

The approach to this would be to list the available CRDs in the cluster and then for each see if there are usages of that CRD

yonahd:
If someone can take this it would be awesome!

hzliangbin:
@yonahd May I have a try？

m-yosefpor:
Oh sorry I forgot to mention here, I implemented the feature already.. So shall I open a PR for this now?

@hzliangbin @yonahd

yonahd:
> Oh sorry I forgot to mention here, I implemented the feature already.. So shall I open a PR for this now?
> 
> @hzliangbin @yonahd

If you already implemented it, open the pr.

hzliangbin:
@yonahd take it

PR #106, opened 2023-10-16
feat: add support to detect unused CRDs

m-yosefpor:
fixes https://github.com/yonahd/kor/issues/103


yonahd:
I reviewed the code on a high level and it looks overall good.(we'll do a proper code review in the next few days hopefully)

The question here is whether presenting the user with individual unused CRDs adds value.

Take `monitoring.coreos.com` as an example, under it there are multiple CRDs e.g.
- Prometheus
- PrometheusRule
- ServiceMonitor

If only one of the above is not used can the user uninstall that individual CRD or is he bound by the higher level of `monitoring.coreos.com`?

Maybe we should be returning only the higher-level that all it's CRDs are not used, WDYT?

m-yosefpor:
> I reviewed the code on a high level and it looks overall good.(we'll do a proper code review in the next few days hopefully)
> 
> The question here is whether presenting the user with individual unused CRDs adds value.
> 
> Take `monitoring.coreos.com` as an example, under it there are multiple CRDs e.g.
> 
>     * Prometheus
> 
>     * PrometheusRule
> 
>     * ServiceMonitor
> 
> 
> If only one of the above is not used can the user uninstall that individual CRD or is he bound by the higher level of `monitoring.coreos.com`?
> 
> Maybe we should be returning only the higher-level that all it's CRDs are not used, WDYT?

It actually really depends.. for example for the monitoring.coreos.com, the prometheus operator neeeds all the CRDs to exist.. but for other project (e.g. argoproj.io) it has many components (argo-workflow, argo-events, argo-cd, argo-appset controller), which all of those CRDs are in the same higher level api group, but they can be deployed independently and each of those use their own set of CRDs in that api group. So I think to cover all such cases, it's better to left the decision on whether he can remove the CRDs or not, to user and we will only report on a per CRD basis.

pkg/kor/kor.go
```diff
@@ -80,6 +82,78 @@ func GetKubeClient(kubeconfig string) *kubernetes.Clientset {
 	return clientset
 }
 
+func GetAPIExtensionsClient(kubeconfig string) *apiextensionsclientset.Clientset {
```
yonahd, line 85: This is very similar code to `GetKubeClient` function
Let's create a function that gets the config and we can create different clientsets based off this function

Same for dynamic client

yonahd (commented):
Looks good overall.
Please rebase and see the comment for making this code cleaner

yonahd:
Also if you can add a couple test for this that would be great!

yonahd:
@m-yosefpor Do you intend to finish this MR?

PR #118, opened 2023-10-26
add initial dashboard

yonahd:


yonahd:
This is basically how it will look
dashboard-screenshot-1

Issue #38, opened 2023-08-29
Discuss: Add Delete Option for Unused Resources


loaynaser3:
I'm a new user of Kor, and I'm already impressed with how great of a tool it is! With that said, 
I'd like to propose an enhancement or new feature for consideration.

**Enhancement Proposal**

addition of a `--delete` option for unused resources in the kor tool. 
This enhancement would provide users with the ability to delete resources that are identified as unused.

**Expected behavior**

The deletion process should be initiated for the unused resources.
The command should provide feedback on the deletion process, indicating which resources are being deleted and their status.

yonahd:
Hi @loaynaser3 
While this would be a great feature, I don't want to put this in the tool.
The main reason is that the tool does not 100% guarantee the resource is unused as you can see here https://github.com/yonahd/kor#supported-resources-and-limitations
And with CRD's we can detect if they are using a resource as well.

If someone wants to use this to delete despite this warning they can import kor and easily get the resources to delete

yonahd:
Reopening.
We might implement this with a safe delete

ozgurcd:
I believe a --dry-run option would be a nice addition at the same time.

yonahd:
> I believe a --dry-run option would be a nice addition at the same time.

Hi @ozgurcd
What do you mean dry run?
Isn't the regular output of the tool a dry run?

Regardless when we put in the delete feature it will be implemented in a safe interactive way

ozgurcd:
Yeah, you're right.

PR #120, opened 2023-10-30
feat: add delete flag

patricktalmeida:
This PR covers discussion on #38.


pkg/kor/delete.go
```diff
@@ -0,0 +1,91 @@
...
+		if !noInteractive {
+			fmt.Printf("Do you want to delete %s %s in namespace %s? (Y/N): ", resourceType, resourceName, namespace)
+			var confirmation string
+			_, err := fmt.Scanf("%s", &confirmation)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Failed to read input: %v\n", err)
+				continue
+			}
+
+			if confirmation != "y" && confirmation != "Y" && confirmation != "yes" {
```
luisdavim, line 76: nit: convert to lower before comparing...

cmd/kor/configmaps.go
```diff
@@ -14,7 +14,7 @@ var configmapCmd = &cobra.Command{
 	Args:    cobra.ExactArgs(0),
 	Run: func(cmd *cobra.Command, args []string) {
 		clientset := kor.GetKubeClient(kubeconfig)
-		if response, err := kor.GetUnusedConfigmaps(includeExcludeLists, clientset, outputFormat, slackOpts); err != nil {
+		if response, err := kor.GetUnusedConfigmaps(includeExcludeLists, clientset, outputFormat, slackOpts, deleteOpts); err != nil {
```
luisdavim, line 17: nit: to avoid having to change all these functions every time we need more data to be passed, we merge the `slackOpts` and `deleteOpts` into an opts struct that would carry all the info, you can even compose it from the existing structs

luisdavim (approved):
LGTM, thanks

PR #121, opened 2023-10-31
Enhanced Dashboard Features: Namespace And Kind Filtering, Resource Visualization.

Yuni-sa:
Added all/multi selection filtering for namespaces and filtering by kind.
Added new panel for displaying all unused resources (pie and bar chart options).
Retained original panels at the bottom with added namespace filter options for comparison.


yonahd (approved):
LGTM!
I'll make small changes and merge it into master

PR #125, opened 2023-11-05
fix: Resolve panic error due to redefined flags in kor

Yuni-sa:
Description:
The issue was causing a panic error when running kor due to redefined flags on root.cmd (lines 58-60). This commit addresses the problem by removing the duplicated flags.

Fixes: #124


yonahd (approved):
Thanks!

PR #135, opened 2023-11-09
grafana dashboard ds selector and repeat kinds

northys:
I wanted to do at least some contribution to this brilliant exporter and I'm good at Grafana so:
* I've updated dashboard to use datasource selector - helpful when you have one grafana with access to multiple prometheus instances
* I've updated the bottom pannels to repeating so you don't have to update the dashboard each time new Kind is exported / supported by kor.
* I did some manual cleanup in the exported dashboard because the oficial Dashboard export option via share placed some internal IDs of our instance
* I don't have grafana with kor datasource without sensitive information so I can't update the image :(

Sadly I have grafana 10 so please try if you can import it to your grafana 9 (maybe with the `requires` change back to 9.4.3 I don't see much special about my changes so it must work)

Data source selector
image

Repeat by variable + variable $kind in query
image

### DIFF
--- a/grafana/dashboard.json
+++ b/grafana/dashboard.json
@@ -1,14 +1,4 @@
 {
-  "__inputs": [
-    {
-      "name": "DS_PROMETHEUS",
-      "label": "prometheus",
-      "description": "",
-      "type": "datasource",
-      "pluginId": "prometheus",
-      "pluginName": "Prometheus"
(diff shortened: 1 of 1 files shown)


yonahd (approved):
LGTM

northys:
Added you on linkedin

PR #140, opened 2023-11-10
chore: remove unused block

slimm609:
removes this block as it is not needed. Cobra handles this before it gets to this condition


yonahd:
@slimm609 Looks good.
Can you just rebase?

PR #149, opened 2023-11-13
fix: fix output for yaml and json, banner

slimm609:
- fix ouput for yaml and json with verbose
- do not show banner on yaml or json


yonahd (approved):
LGTM

Issue #123, opened 2023-11-04
Feature: Add interactive Label as used


yonahd:
When a user gets prompted to delete a resource and chooses not to delete, offer to label the resource as used using this label feature https://github.com/yonahd/kor#ignore-resources

patricktalmeida:
Commenting for task assignment

PR #150, opened 2023-11-13
feat: add inUse label prompt for non deleted resources

patricktalmeida:
This PR covers issue #123

### DIFF
--- a/pkg/kor/delete.go
+++ b/pkg/kor/delete.go
@@ -4,8 +4,15 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"reflect"
 	"strings"
 
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1beta1 "k8s.io/api/networking/v1beta1"
+	policyv1beta1 "k8s.io/api/policy/v1beta1"
+	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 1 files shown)


yonahd:
Maybe we can have some generic function something like 
```
func FlagResource(clientset kubernetes.Interface, namespace, resourceType, resourceName string) error {
	// Get the resource dynamically based on the resource type
	resource, err := getResource(clientset, namespace, resourceType, resourceName)
	if err != nil {
		return err
	}

	// Update the labels using reflection
	labelField := reflect.ValueOf(resource).Elem().FieldByName("Labels")
	if labelField.IsValid() {
		labels := labelField.Interface().(map[string]string)
		if labels == nil {
			labels = make(map[string]string)
		}
		labels["kor/used"] = "true"
		labelField.Set(reflect.ValueOf(labels))
	} else {
		return fmt.Errorf("unable to set labels for resource type: %s", resourceType)
	}

	// Update the resource
	_, err = updateResource(clientset, namespace, resourceType, resource)
	return err
}

```

And just have a switch case for getting and updating e.g.(and similar function for updating)
```
func getResource(clientset kubernetes.Interface, namespace, resourceType, resourceName string) (interface{}, error) {
	switch resourceType {
	case "ConfigMap":
		return clientset.CoreV1().ConfigMaps(namespace).Get(context.TODO(), resourceName, metav1.GetOptions{})
	case "Secret":
		return clientset.CoreV1().Secrets(namespace).Get(context.TODO(), resourceName, metav1.GetOptions{})
	....
}
```

yonahd (approved):
LGTM!

Issue #167, opened 2023-11-28
Feature: Add reason for marking as unused


yonahd:
As a preparation for 
https://github.com/yonahd/kor/issues/117
And
https://github.com/yonahd/kor/issues/107
Add the reason kor is considering the resource unused.
E.g.

- Unused
- Marked with unused label
- In termination waiting for finalizers to be removed.

Issue #117, opened 2023-10-24
Support resources waiting for finalizers to delete


yonahd:
Do something similar to  https://github.com/ibuildthecloud/finalizers

Yuni-sa:
can I be assigned to this issue?

yonahd:
@Yuni-sa Take into account this https://github.com/yonahd/kor/issues/38
They might be slightly conflicting

PR #172, opened 2023-11-30
Add finalizer subcommand

Yuni-sa:
Iterates over all types of namespaced resources in the cluster (Including Custom Resources) check if they are in a terminating state by looking if they have a finalizer and a deletion timestamp, deletion of those resources is also supported by patching out the finalizer.

part of #117 

this command is not added to the "all" command for future plans implementing a separate function for the supported resources after #167


yonahd:
@Yuni-sa is this ready for review?

pkg/kor/finalizers.go
```diff
@@ -0,0 +1,128 @@
...
+
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to process resources waiting for finalizers: %v\n", err)
+	}
+
+	allDiffs := make(map[string][]string)
+
+	for namespace, resourceType := range pendingDeletionDiffs {
+		if slices.Contains(namespaces, namespace) {
+			for gvr, resourceDiff := range resourceType {
```
yonahd, line 100: What is gvr?

yonahd (commented):
Overall it looks good.
Can we support this in multi and all and merge the namespaces?
Maybe we can do it that way and then put finalizers as include finalizers flag or something like that?

WDYT?

pkg/kor/finalizers.go
Yuni-sa: schema of group version resource, resource stands for the name of the resource in the same format of the api endpoint which is lowercase plural.
for example: Kind- `Pod`, Resource-`pods`

Yuni-sa:
> Overall it looks good. Can we support this in multi and all and merge the namespaces? Maybe we can do it that way and then put finalizers as include finalizers flag or something like that?
> 
> WDYT?

Not sure if there is a good implementation of this without #167 , after reasoning will be added I will create a new pr that will include it in all and multi.

if we add this now it would probably all be deleted/changed in the next pr, I think the current implementation of using `kor finalizers` which also supports all of the filters options is fine for now.

yonahd:
> > Overall it looks good. Can we support this in multi and all and merge the namespaces? Maybe we can do it that way and then put finalizers as include finalizers flag or something like that?
> > WDYT?
> 
> Not sure if there is a good implementation of this without #167 , after reasoning will be added I will create a new pr that will include it in all and multi.
> 
> if we add this now it would probably all be deleted/changed in the next pr, I think the current implementation of using `kor finalizers` which also supports all of the filters options is fine for now.

Ok.
Can you just add the command in the Readme and then we can merge it

yonahd (approved):
LGTM
Thanks!

Issue #170, opened 2023-11-29
Feature: Find pods in evicted state


yonahd:
Find pods that are in the state evicted and report them as unused

yonahd:
Assigned. Thanks!

PR #173, opened 2023-12-03
feat: find evicted pods

Nezz7:
Find evicted pods: pods with status = `Failed` and Reason = `Evicted`.
resolve [#170 ]( https://github.com/yonahd/kor/issues/170)


yonahd:
> it would be better to output the `PodStatus.Message` when deleting `evicted` pods. wdyt?

We currently don't do that for any resource. But this might be something we can do as part of https://github.com/yonahd/kor/issues/167

pkg/kor/delete.go
```diff
@@ -55,6 +55,9 @@ func DeleteResourceCmd() map[string]func(clientset kubernetes.Interface, namespa
 		"PV": func(clientset kubernetes.Interface, namespace, name string) error {
 			return clientset.CoreV1().PersistentVolumes().Delete(context.TODO(), name, metav1.DeleteOptions{})
 		},
+		"Pod": func(clientset kubernetes.Interface, namespace, name string) error {
```
yonahd, line 58: Please also add Pod the flagResource and getResource

yonahd (approved):
LGTM

PR #176, opened 2023-12-08
Fix exporterInterval variable.

MostovykVlad:
exporterInterval variable should be added to values.yaml without quotes and should be added to env with quotes.


Yuni-sa:
Nice catch!

yonahd (approved):
LGTM
Thanks!

PR #180, opened 2023-12-17
feat: add replicaset cleaner

cindyliu-tec:


pkg/kor/replicaset.go
```diff
@@ -0,0 +1,85 @@
...
+			continue
+		}
+		// checks if the resource's age (measured from its last modified time) matches the included criteria
+		// specified by the filter options.
+		if included, _ := HasIncludedAge(replicaSet.CreationTimestamp, filterOpts); !included {
+			continue
+		}
+
+		// if the replicaSet is specified 0 replica and current available & ready & fullyLabeled replica count is all 0, think the replicaSet is completed
+		if *replicaSet.Spec.Replicas == 0 && replicaSet.Status.AvailableReplicas == 0 && replicaSet.Status.ReadyReplicas == 0 && replicaSet.Status.FullyLabeledReplicas == 0 {
```
yonahd, line 38: What is `FullyLabeledReplicas`?

cindyliu-tec: fullyLabeledReplicas is the number of pods that have labels matching the labels of the pod template of the replicaSet. if a replicaSet desire to scale pods to 0, when  it complete it's work, available & ready & fullyLabeled should all be 0, and barely grow again.

yonahd: I see some unused replicasets do not contain all these fields. This will likely cause an error

cindyliu-tec: yes, i see this too, so i check all this property, that ensure do not make wrong deletion

yonahd: @cindyliu-tec But what will happen when we run this check and the field doesn't exist?
Kor will likely crash

cindyliu-tec: as golang language design, a struct field  without assignment will default assigned the zero value of it's type. so if  the field doesn't exist, it's value would be 0. when a replicaSet is still active, it must has those fields, and impossible be all 0.

cindyliu-tec: as golang language design, a struct field  without assignment will default assigned the zero value of it's type. so if  the field doesn't exist, it's value would be 0. when a replicaSet is still active, it must has those fields, and impossible be all 0.

yonahd (approved):
LGTM

Issue #187, opened 2024-01-17
Feature: Add include-labels filter flag


philippe-granet:
**Is your feature request related to a problem? Please describe.**
I want to only clean secrets that have certains labels

**Describe the solution you'd like**
Selector to filter label, Example: --include-labels key1=value1,key2=value2.

yonahd:
This should be fairly simple to add as we already support the exclude labels.
@philippe-granet would you like to attempt this?

philippe-granet:
Sorry, I've never coded in Go and I don't really have time available at the moment

Forget-C:
I want to add a filtering interface to implement unified resource filtering logic.
currently, if you want to add a filtering method, it needs to be processed in each resource implementation.
I want to change to this mode:
```go
func ProcessNamespaceDeployments(clientset kubernetes.Interface, namespace string, filterOpts *FilterOptions) ([]string, error) {
	deploymentsList, err := clientset.AppsV1().Deployments(namespace).List(context.TODO(), metav1.ListOptions{})
	if err != nil {
		return nil, err
	}

	var deploymentsWithoutReplicas []string

	for _, deployment := range deploymentsList.Items {

        filterFunc := func(obj interface) bool {
            deployment := obj.(*appsv1.Deployment)
            if *deployment.Spec.Replicas == 0 {
			    return true
		    }
            return false
        }
        // Run will run the registered filter method
		if pass := filter.AddFilter(filterFunc).Run();pass{
            deploymentsWithoutReplicas = append(deploymentsWithoutReplicas, deployment.Name)
        }
	}

	return deploymentsWithoutReplicas, nil
}
func someInit() {
    filter := NewFilter()
    filter = filter.AddFilter(func(obj interface{}) bool {
        return !HasExcludedLabel(deployment.Labels, filterOpts.ExcludeLabels)
    })
    filter = filter.AddFilter(func(obj interface{}) bool {
        return HasIncludedAge(deployment.CreationTimestamp, filterOpts)
    })
    filter = filter.AddFilter(func(obj interface{}) bool {
        deployment := obj.(*appsv1.Deployment)
        if deployment.Labels["kor/used"] == "true" {
			return true
		}
    })
}
```
The filter object implements unified filtering logic, and the different filtering methods call `filter.AddFilter(filterFunc).Run()`.

yonahd:
[...]

PR #194, opened 2024-01-25
Refactor some codes, add filter objects, and add support for lable filtering...

Forget-C:
refactor&feat&fix&style:
    1. Unify parameter types and divide them into two categories: normal parameters `kor.Opts` and filter parameters `filters.Options`.
    2. `includeExcludeLists` no longer participates in `cmd args` binding, and is taken over by `filters.Options`.
    3. The initialization operations of `includeNamespace` and `excludeNamespace` are taken over by `filters.Options` and will only be executed once.
    4. Add filters.Framework object to uniformly process filtering logic.
    5. All currently supported resources are supported by the `kor/used=true` tag.
    6. Fix the problem that filterOptions.Validate does not take effect
    7. Support includeable parameters

The unit tests have passed, but due to the large scope of the changes, more tests are needed.


Forget-C:
The original purpose was to abstract the `filter framework` to facilitate the addition of filter parameters. Other changes are necessary to achieve this goal. #187 
now, If we want to add a filter, we only need the following steps：
#### create filter func
```go
// FilterFunc is a filter that is a function
// If the resource is legal, return true
// example:
// deployment.Spec.Replicas > 0; return true
// meta.GetLabels()["kor/used"] == "true"; return true
type FilterFunc func(object runtime.Object, opts *Options) bool
func TestFilter(object runtime.Object, opts *Options) bool {
	if meta, ok := object.(metav1.Object); ok {
		if meta.GetLabels()["kor/used"] == "true" {
			return true
		}
	}
	return false
}
``` 
#### Set a name
```go
const  TestFilterName = "test"
```
#### Add to registry
```go
func NewDefaultRegistry() Registry {
	return Registry{
		LabelFilterName:    LabelFilter,
		AgeFilterName:      AgeFilter,
		KorLabelFilterName: KorLabelFilter,
                TestFilterName: TestFilter,
	}
}
```

cmd/kor/root.go
```diff
@@ -50,19 +49,26 @@ func Execute() {
...
 	if err := rootCmd.Execute(); err != nil {
 		fmt.Fprintf(os.Stderr, "Error while executing your CLI '%s'", err)
 		os.Exit(1)
 	}
 }
 
-func addFilterOptionsFlag(cmd *cobra.Command, opts *kor.FilterOptions) {
-	cmd.PersistentFlags().StringVarP(&opts.ExcludeLabels, "exclude-labels", "l", opts.ExcludeLabels, "Selector to filter out, Example: --exclude-labels key1=value1,key2=value2.")
+func addFilterOptionsFlag(cmd *cobra.Command, opts *filters.Options) {
+	cmd.PersistentFlags().StringVarP(&opts.ExcludeLabels, "exclude-labels", "l", opts.ExcludeLabels, "Selector to filter out, Example: --exclude-labels key1=value1,key2=value2. If --include-labels is set, --exclude-labels will be ignored.")
```
luisdavim, line 68: why not allow both to exist together? maybe I have a bunch of objects that generally I want to include but there are some exceptions I want to exclude with some extra set of labels.

pkg/kor/deployments.go
```diff
@@ -9,32 +9,22 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+
+	"github.com/yonahd/kor/pkg/filters"
 )
 
-func ProcessNamespaceDeployments(clientset kubernetes.Interface, namespace string, filterOpts *FilterOptions) ([]string, error) {
-	deploymentsList, err := clientset.AppsV1().Deployments(namespace).List(context.TODO(), metav1.ListOptions{})
+func ProcessNamespaceDeployments(clientset kubernetes.Interface, namespace string, filterOpts *filters.Options) ([]string, error) {
+	deploymentsList, err := clientset.AppsV1().Deployments(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: filterOpts.IncludeLabels})
```
luisdavim, line 17: this makes a lot of sense so we don't fetch unwanted resources from the API but we could still used the `ExcludeLabels` below to filter the returned list.

pkg/filters/filters.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+	return false
+}
+
+// LabelFilter is a filter that filters out resources by label
+func LabelFilter(object runtime.Object, opts *Options) bool {
+	if meta, ok := object.(metav1.Object); ok {
+		if has, err := HasExcludedLabel(meta.GetLabels(), opts.ExcludeLabels); err == nil {
+			return has
```
luisdavim, line 32: if the object matches the exclusion selector shouldn't it be filtered? shouldn't this be `return !has`?

pkg/filters/options.go
```diff
@@ -0,0 +1,147 @@
...
+//   - It does not have any labels that match the ExcludeLabels flag. The ExcludeLabels flag supports '=', '==', and '!=' operators,
+//     and multiple label pairs can be separated by commas. For example, -l key1=value1,key2!=value2.
+type Options struct {
+	// OlderThan is the minimum age of the resources to be considered unused
+	OlderThan string
+	// NewerThan is the maximum age of the resources to be considered unused
+	NewerThan string
+	// ExcludeLabels is a label selector to exclude resources with matching labels
+	// IncludeLabels conflicts with it, and when setting IncludeLabels, ExcludeLabels is ignored and set to empty
+	ExcludeLabels string
```
luisdavim, line 32: nit: I think it would be more intuitive to call this something like `ExcludeSelector` to make it clearer it's a selector string.

pkg/filters/options.go
```diff
@@ -0,0 +1,147 @@
...
+	// IncludeLabels is a label selector to include resources with matching labels
+	IncludeLabels string
```
luisdavim, line 34: same as above about the name.

pkg/filters/options.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// Modify modifies the options
+func (o *Options) Modify() {
+	o.modifyLabels()
+}
+
+// Namespaces returns the namespaces, only called once
+func (o *Options) Namespaces(clientset kubernetes.Interface) []string {
+	o.once.Do(func() {
```
luisdavim, line 95: nit: instead of `once` we could jut check if `o.namespce` is `nil`.

luisdavim (commented):
Overall LGTM, left a few comments in the code.

Thanks.

cmd/kor/root.go
Forget-C: 1. I want to be consistent with the current ns processing. 
2. both hoping and hoping for a less effective approach.
ps: The current way of initiating requests by namespace is also very inefficient, causing unnecessary pressure on the apiserver, although in many cases the number of ns is not large.

pkg/kor/deployments.go
Forget-C: I think this is the same problem as the above: Do exclude and include conflict?

pkg/filters/options.go
Forget-C: okay ， this good. I want to be consistent with the prompts in the command.  `cmd --exclude-labels` --> `cmd --exclude-selector` . (--selector keep similar parameters to kubectl)

pkg/filters/options.go
Forget-C: Checking ns saves an extra assignment. The advantage of once is that it is only executed once and does not require additional checks. However, checking ns will still make multiple requests to the server in some cases, such as when the server is busy during a period of time.（Adding additional judgments increases complexity）

pkg/kor/deployments.go
yonahd: I'd like to consider them conflicting

pkg/filters/options.go
yonahd: Wouldn't we have the same issue if the server is busy even with the once implementation?

Forget-C: There are still some differences. When checking multiple resources at the same time (deploy, rs...)
once: No matter whether there is a problem or not, it will only be accessed once.
check: If the deploy check fails and the namespace string is empty, the next resource will continue to request services.
Supportive tools should have minimal impact on service.

PR #199, opened 2024-02-02
Feat: added support for unused DaemonSets

doronkg:
This PR adds a new functionality to `kor` by finding unused DaemonSets.
I've considered DaemonSets to be unused, if the following condition is met:
```
status:
  currentNumberScheduled: 0
```

Meaning the number of nodes that should be running at least 1 daemon pod and are supposed to run the daemon pod, is equal to zero.
More information about DaemonSets' status fields could be found in these [docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#daemonsetstatus-v1-apps).


doronkg:
Conflicts might raise in files `pkg/kor/delete.go` (#201) & `charts/kor/templates/role.yaml` (#202).
Should be taken into account when merging.

pkg/kor/create_test_resources.go
```diff
@@ -321,4 +321,33 @@ func CreateTestReplicaSet(namespace, name string, specReplicas *int32, status *a
...
+}
+
+func CreateTestDaemonSet(namespace, name string, labels map[string]string, status *appsv1.DaemonSetStatus) *appsv1.DaemonSet {
+	return &appsv1.DaemonSet{
+		ObjectMeta: v1.ObjectMeta{
+			Namespace: namespace,
+			Name:      name,
+			Labels:    labels,
+		},
+		Spec: appsv1.DaemonSetSpec{
```
yonahd, line 333: nit: do we need the spec part of the daemonset?

yonahd (commented):
Lot great overall small comment

pkg/kor/create_test_resources.go
doronkg: Was indeed excessive, test ran successfully without it. Removed.
Thanks

yonahd (approved):
LGTM
Thanks

PR #201, opened 2024-02-03
Fix: refactored mismatching deletion resource types

### LINKED ISSUE #200: Broken Role/SA/STS Deletion
**Describe the bug**
When trying to delete resources of types [role|serviceaccount|statefulset], the resource-type is wrongly passed as the name of the requested instance to delete, and the deletion itself fails.

**To Reproduce**
Run `kor [role|serviceaccount|statefulset] --delete`.

**Expected behavior**
The resources would be deleted.

**Screenshots**
image

Kor version: 0.3.3

doronkg:
In cases where the resource type name mismatches in the following funcs, its deletion functionality will break:
```
# pkg/kor/<resource>s.go

func GetUnused<Resource>s()
	if opts.DeleteFlag {
          if diff, err = DeleteResource(diff, clientset, namespace, "<Resource>", opts.NoInteractive); err != nil {
          
# pkg/kor/delete.go

func DeleteResourceCmd()
        "<Resource>": func(clientset kubernetes.Interface, namespace, name string) error {
```

I've fixed the following mismatches in order to resolve #200:
- `delete.go` -> changed `Roles` to `Role` to match passed resource-type in `roles.go`.
- `serviceaccountss.go` -> changed `Serviceaccount` to `ServiceAccount`.
- `statefulsets.go` -> changed `Statefulset` to `StatefulSet`.

### DIFF
--- a/pkg/kor/delete.go
+++ b/pkg/kor/delete.go
@@ -45,7 +45,7 @@ func DeleteResourceCmd() map[string]func(clientset kubernetes.Interface, namespa
 		"PDB": func(clientset kubernetes.Interface, namespace, name string) error {
 			return clientset.PolicyV1beta1().PodDisruptionBudgets(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
 		},
-		"Roles": func(clientset kubernetes.Interface, namespace, name string) error {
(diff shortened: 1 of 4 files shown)


yonahd (approved):
Good catches
Thanks!

PR #202, opened 2024-02-03
Refactor: added missing resources and separated cluster-scoped resources

doronkg:
The following resources were missing from the `kor` Helm chart RBAC:
- jobs
- replicasets
- persistentvolumes
- customresourcedefinitions

In addition, the following resources are cluster-scoped and won't have any effect when listed under `read-resources-role`, only under `read-resources-clusterrole`:
- namespaces
- clusterroles
- clusterrolebindings

### DIFF
--- a/charts/kor/templates/role.yaml
+++ b/charts/kor/templates/role.yaml
@@ -10,7 +10,6 @@ rules:
     resources:
       - pods
       - configmaps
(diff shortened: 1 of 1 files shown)


yonahd (approved):
LGTM
Thanks

PR #210, opened 2024-02-23
Feat: Added version subcommand

doronkg:
Added `version` subcommand for users to view `kor` version not through querying commands (banner display).

```
$ kor version
kor version: v<release>
```


yonahd (approved):
LGTM

Issue #218, opened 2024-03-04
Feature: In-cluster exporter configuration 


hrytsai:
**Is your feature request related to a problem? Please describe.**
We run kor outside of the cluster with option provided below and got expected result: list of unused configmaps,secrets and pvc's in namespace
`kor configmap,secret,persistentvolumeclaim -n kor-test-namespace`
And now we want to reproduce this scenario but in cluster with export mode.

**Describe the solution you'd like**
Possibility to configure kor to export prometheus metric only from 1 namespace and reduce list of resources, for example only for: configmap,secret,persistentvolumeclaim in "In cluster" mode ?
Something like: 
`kor exporter -n columbus-data-platform-dev -r configmap,secret,persistentvolumeclaim (List of included resources, by default it can be: all)`

**Describe alternatives you've considered**

**Additional context**
We want use kor in environment where each project is isolated in own Project\Namespace in OpenShift 4.x platform and should keep tracking unused resources separately.

Yuni-sa:
@yonahd can I be assigned to this?

yonahd:
@Yuni-sa I believe namespaces are already supported

Yuni-sa:
@yonahd ,Yes it seems they do work as intended, I think I would add the functionality using GetUnusedMulti().

what do you think would be better:

```kor exporter svc,ing,pod -n namespace```
or 
``` kor exporter -r svc,ing,pod -n namespace```

yonahd:
I actually think the proper setup may be

`kor svc, ing -n namespace -exporter `

But using our current code we can do it with -r

PR #229, opened 2024-03-25
Feat: Add support for multiple resources in Exporter function

Yuni-sa:
solves #218
allows you to select which types of unused resources you want to find:
``kor exporter --resources svc,deploy,ing -n namespace`` , can also use `-r` for short.

if resources are not specified it defaults to all resources.

### DIFF
--- a/pkg/kor/exporter.go
+++ b/pkg/kor/exporter.go
@@ -80,8 +80,9 @@ func exportMetrics(filterOptions *filters.Options, clientset kubernetes.Interfac
 }
 
 func getUnusedResources(filterOptions *filters.Options, clientset kubernetes.Interface, apiExtClient apiextensionsclientset.Interface, dynamicClient dynamic.Interface, outputFormat string, opts Opts, resourceList []string) (string, error) {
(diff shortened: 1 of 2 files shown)


pkg/kor/exporter.go
```diff
@@ -77,3 +78,10 @@ func exportMetrics(filterOptions *filters.Options, clientset kubernetes.Interfac
 		}
 	}
 }
+
+func getUnusedResources(filterOptions *filters.Options, clientset kubernetes.Interface, apiExtClient apiextensionsclientset.Interface, dynamicClient dynamic.Interface, outputFormat string, opts Opts, resourceList []string) (string, error) {
+	if len(resourceList) > 0 {
```
yonahd, line 83: edge case here might be when a user enters all

yonahd (commented):
Looks good overall. small comment

pkg/kor/exporter.go
Yuni-sa: good catch!

yonahd (approved):
LGTM
Thanks!

PR #233, opened 2024-04-03
Show unused reason 

yonahd:


yonahd:
@luisdavim Can you take a look at this idea
For ticket https://github.com/yonahd/kor/issues/167

doronkg:
I suggest renaming `--print-reason` flag to `--show-reason` in similarity to kubectl flags.

pkg/kor/all.go
```diff
@@ -23,6 +23,11 @@ type ResourceDiff struct {
 	diff         []string
 }
 
+type ResourceDiff2 struct {
```
luisdavim, line 26: why a new type?

pkg/kor/all.go
```diff
@@ -23,6 +23,11 @@ type ResourceDiff struct {
...
+	resourceType string
+	diff         interface{}
```
luisdavim, line 28: I think that instead of ising an interface why not always use the `ResourceInfo` and depending on the flag, set/display the reson or not...

pkg/kor/all.go
yonahd: This is temporary until I switch them all

pkg/kor/all.go
yonahd: Hmm. This might be significantly cleaner. I'll give it a try

Issue #235, opened 2024-04-17
Map false unused resources: Kind


yonahd:
Find all default resources in a kind cluster that kor flags as unused.
List them
Create an issue to exclude them
Bonus: Open merge request with the exclusions
(example can be found here)

yonahd:
I have docker.io-hostpath in kind (macos)

doronkg:
As mentioned in https://github.com/yonahd/kor/pull/241#issuecomment-2077220002, multiple resources are similar between `minikube` & `kind`. 

@yonahd should I append the differences to the existing exclusion lists for each resource, or await #253?

doronkg:
> I have docker.io-hostpath in kind (macos)

I can perform checks both for Windows & Linux installations, but don't have MacOS, so I'll add that.
Have we performed `minikube` checks for all these distributions as well? @cecobask

cecobask:
Hi @doronkg,

I tested the minikube changes on a MacOS only. Might have to investigate further in a separate ticket. Feel free to open up an issue and I can pick it up soon.

doronkg:
> Hi @doronkg,
> 
> 
> 
> I tested the minikube changes on a MacOS only. Might have to investigate further in a separate ticket. Feel free to open up an issue and I can pick it up soon.

I can take a look on the other OSs for `minikube` and see if additional resources are required for exclusion.

I'd appreciate if you can assist me with checking `kind` on MacOS after I'll submit the PR to see if more resources should be excluded.

yonahd:
> > Hi @doronkg,
> > I tested the minikube changes on a MacOS only. Might have to investigate further in a separate ticket. Feel free to open up an issue and I can pick it up soon.
> 
> I can take a look on the other OSs for `minikube` and see if additional resources should be excluded.
> 
> I'd appreciate if you can assist me with checking `kind` on MacOS after I'll submit the basic PR to see if more resources should be excluded.

I'll make sure you get it all from macos

doronkg:
> As mentioned in https://github.com/yonahd/kor/pull/241#issuecomment-2077220002, multiple resources are similar between `minikube` & `kind`. 
> 
> 
> 
> @yonahd should I append the differences to the existing exclusion lists for each resource, or await #253?

@yonahd

yonahd:
> > As mentioned in [#241 (comment)](https://github.com/yonahd/kor/pull/241#issuecomment-2077220002), multiple resources are similar between `minikube` & `kind`.
[...]

PR #241, opened 2024-04-18
refactor: map false minikube unused resources

### LINKED ISSUE #238: Map false unused resources: minikube
Find all default resources in a minikube cluster that kor flags as unused.
List them
(example can be found here  https://github.com/yonahd/kor/blob/main/pkg/kor/serviceaccounts.go#L16)

cecobask:
Fixes #238

### DIFF
--- a/pkg/kor/configmaps.go
+++ b/pkg/kor/configmaps.go
@@ -14,9 +14,39 @@ import (
 	"github.com/yonahd/kor/pkg/filters"
 )
 
-var exceptionconfigmaps = []ExceptionResource{
-	{ResourceName: "aws-auth", Namespace: "kube-system"},
-	{ResourceName: "kube-root-ca.crt", Namespace: "*"},
+var exceptionConfigMaps = []ExceptionResource{
+	{
+		ResourceName: "aws-auth",
+		Namespace:    "kube-system",
+	},
+	{
+		ResourceName: "kube-root-ca.crt",
+		Namespace:    "*",
+	},
+	{
+		ResourceName: "extension-apiserver-authentication",
+		Namespace:    "kube-system",
+	},
+	{
+		ResourceName: "kube-apiserver-legacy-service-account-token-tracking",
+		Namespace:    "kube-system",
+	},
+	{
+		ResourceName: "kubeadm-config",
+		Namespace:    "kube-system",
+	},
+	{
+		ResourceName: "kubelet-config",
+		Namespace:    "kube-system",
+	},
+	{
+		ResourceName: "kubernetes-dashboard-settings",
+		Namespace:    "kubernetes-dashboard",
+	},
+	{
+		ResourceName: "cluster-info",
+		Namespace:    "kube-public",
+	},
 }
 
 func retrieveUsedCM(clientset kubernetes.Interface, namespace string) ([]string, []string, []string, []string, []string, error) {
@@ -75,7 +105,7 @@ func retrieveUsedCM(clientset kubernetes.Interface, namespace string) ([]string,
 		}
 	}
 
-	for _, resource := range exceptionconfigmaps {
+	for _, resource := range exceptionConfigMaps {
 		if resource.Namespace == namespace || resource.Namespace == "*" {
 			volumesCM = append(volumesCM, resource.ResourceName)
 		}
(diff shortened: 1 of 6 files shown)


pkg/kor/configmaps.go
```diff
@@ -126,7 +156,13 @@ func processNamespaceCM(clientset kubernetes.Interface, namespace string, filter
...
 
 	var usedConfigMaps []string
-	slicesToAppend := [][]string{volumesCM, envCM, envFromCM, envFromContainerCM, envFromInitContainerCM}
+	slicesToAppend := [][]string{
+		volumesCM,
+		envCM,
+		envFromCM,
+		envFromContainerCM,
+		envFromInitContainerCM,
+	}
```
cecobask, line 165: Autoformat by IDE

pkg/kor/kor.go
```diff
@@ -123,10 +123,16 @@ func FormatOutput(namespace string, resources []string, resourceType string, opt
 
 	var buf bytes.Buffer
 	table := tablewriter.NewWriter(&buf)
-	table.SetHeader([]string{"#", "Resource Name"})
+	table.SetHeader([]string{
+		"#",
+		"Resource Name",
+	})
```
cecobask, line 129: Autoformat by IDE

pkg/kor/configmaps.go
yonahd: Yeah. We need to enforce it in the CI.
I'll open a task for it

pkg/kor/kor.go
```diff
@@ -3,18 +3,18 @@ package kor
 import (
 	"bytes"
 	"fmt"
-	"os"
```
yonahd, line 6: The imports should be:
- standard library modules
- 3rd party modules
- current project modules

pkg/kor/storageclasses.go
```diff
@@ -14,6 +14,13 @@ import (
 	"github.com/yonahd/kor/pkg/filters"
 )
 
+var exceptionStorageClasses = []ExceptionResource{
```
yonahd, line 17: Is this a false positive? Or do you not have any use of the storage class

pkg/kor/configmaps.go
yonahd: Strange. Intellij doesn't reformat to this.

pkg/kor/kor.go
cecobask: I actually didn’t touch the imports. This must also be part of the automatic changes by my IDE.

I think golangci-lint had a linter for this exact problem.

pkg/kor/storageclasses.go
cecobask: I believe this storage class should be treated as false positive as it comes with the vanilla configuration of minikube.

pkg/kor/configmaps.go
cecobask: There might be some differences in how IntelliJ and GoLand work when it comes to formatting. I’m using the default code style configuration, no custom tweaks there.

doronkg:
This PR also covers #235 almost completely.
I've created a default `kind` cluster and ran `kor all` right away. 
All the unused ClusterRoles, ConfigMaps, Secrets & StorageClasses are mapped in this PR, but for one ClusterRole.

The ClusterRole `system:persistent-volume-provisioner` is created in both distros, but by default:
- In `minikube` it is used by the `storage-provisioner` ClusterRoleBinding.
- In `kind` it is unused by any RB/CRB.

@yonahd would you prefer that I'll submit a different PR for that or rather should we include the single difference in this PR?

yonahd:
> This PR also covers #235 almost completely. I've created a default `kind` cluster and ran `kor all` right away. All the unused ClusterRoles, ConfigMaps, Secrets & StorageClasses are mapped in this PR, but for one ClusterRole.
> 
> The ClusterRole `system:persistent-volume-provisioner` is created in both distros, but by default:
> 
> * In `minikube` it is used by the `storage-provisioner` ClusterRoleBinding.
> * In `kind` it is unused by any RB/CRB.
> 
> @yonahd would you prefer that I'll submit a different PR for that or rather should we include the single difference in this PR?

@doronkg there are not some docker related configmaps specific to kind?

doronkg:
> @doronkg there are not some docker related configmaps specific to kind? 

All the ConfigMaps are the same between `kind` & `minikube` on default installations.

cecobask:
@yonahd, I rebased my branch with the latest changes and fixed the lint issues.
Please, take a look when you get a chance and let me know if something else is left to be done here 👍

pkg/kor/kor.go
cecobask: Just an update on this comment:
The [gci](https://golangci-lint.run/usage/linters/#gci) linter will take care of this now. Every time the linter runs it will check if the package import order is valid and fail the checks when it isn't.

yonahd (approved):
Lgtm
Thanks!

PR #247, opened 2024-04-22
feat(image): build multi-arch image

isindir:
* Update couple of plugins
* Setup and build multi-arch docker image, add arm64 arch


yonahd (approved):
Lgtm
Thanks!

PR #249, opened 2024-04-28
feat(namespaces): add empty namespace detection and removal

isindir:

This PR finds empty and non-default kubernetes namespaces, and if instructed removes these from the cluster.

## PR Checklist

- [x] This PR adds new code
- [ ] This PR adds docs

## GitHub Issue

Closes [#92]

## Notes for your reviewers


isindir:
fixes #92

yonahd:
> fixes #92
> 
> This PR is still WIP, as it lacks parallel processing of the namespaces as well as unit tests, could you please review if in principle it is what you'd be happy to accept ?

Hey @isindir
I'll take a look at the code in the next couple of days

doronkg:
Hi @isindir,
In order for the new feature to be completed, several additional files should be modified accordingly.

Take a look at https://github.com/yonahd/kor/blob/main/CONTRIBUTING.md#repository-structure.

yonahd:
> fixes #92
> 
> This PR is still WIP, as it lacks parallel processing of the namespaces as well as unit tests, could you please review if in principle it is what you'd be happy to accept ?

This is definitely a good direction(this is not a simple one).
Definitely need to review this thoroughly but in principal this is good

pkg/kor/namespaces.go
```diff
@@ -0,0 +1,221 @@
...
+					continue
+				}
+				if gr.GVR.Resource == "events" {
+					continue
+				}
+				// User specified resource type ignore list
+				if ignoreResourceType(gr.GVR.Resource, filterOpts.IgnoreResourceTypes) {
+					continue
+				}
+				genResources = append(genResources, gr)
```
yonahd, line 104: Why are we collecting and not returning false?

isindir: This one is to ignore user specified resource types, for example we use cluster level externalsecrets, which are creating secrets in each namespace, where we consider namespace to be empty if there are only secrets left in the namespace. One can use:
```
kor ns --ignore-resource-types secrets,configmaps
```
that should list any namepsaces where only cms and secrets are left (default resources and events are ignored as well).

isindir:
> Hi @isindir, In order for the new feature to be completed, several additional files should be modified accordingly.
> 
> Take a look at https://github.com/yonahd/kor/blob/main/CONTRIBUTING.md#repository-structure.

Helm chart I think needs extra changes:
* at the moment it is not possible to configure `kor` in a deployment for example to automatically remove any orphaned resources - rbac lacks such permissions.
* I think it would be more useful to split cronjobs so that several such jobs are configured with different schedules (may become handy when removing some resources and then namespaces in a separate run.

I'm looling to work on unit tests if time allows these weekends.

pkg/kor/namespaces.go
yonahd: From what I see we check if this list is longer than 0 to determine if the namespace is unused. Why continue collecting these or iterating after we find one?

isindir: true, I'll rework this function to return true/false - as an answer to "resources found?" . This should speed things up significantly.

isindir: @yonahd fixed this

doronkg:
> Helm chart I think needs extra changes:
> 
> * at the moment it is not possible to configure `kor` in a deployment for example to automatically remove any orphaned resources - rbac lacks such permissions.

Correct, currently the deployment is only used to export metrics, hence read-only RBAC, but we can consider going that way.
> * I think it would be more useful to split cronjobs so that several such jobs are configured with different schedules (may become handy when removing some resources and then namespaces in a separate run.

Interesting, especially when scanning unused resources in high-scaled clusters that might take more time than the interval set.

We can continue with progress on both comments in separated issues/discussions (or in our Discord channel) as they are out-of-scope for this PR, but should be dippen into. I'd like to futher discuss them.

Referring to https://github.com/yonahd/kor/blob/main/CONTRIBUTING.md#repository-structure was made to make sure you modify all required files, as adding any new unused resource support should address them.

isindir:
@doronkg , 

> Referring to https://github.com/yonahd/kor/blob/main/CONTRIBUTING.md#repository-structure was made to make sure you modify all required files, as adding any new unused resource support should address them.

Thank you, I think I covered most of the files except helm chart and some unit tests, atm I'm trying to figure out if I can use fake clients to reliably test a feature I'm implementing. I deliberately left chart untouched as the change already looks very big. 

btw, for chart - it will be better to have fine-tuned RBAC - possibility to enable only those read/write/per namespace permissions per feature (resource).

pkg/kor/namespaces.go
```diff
@@ -0,0 +1,221 @@
...
+func ignoreResourceType(resource string, ignoreResourceTypes []string) bool {
+	for _, ignoreType := range ignoreResourceTypes {
+		if resource == ignoreType {
+			return true
+		}
+	}
+	return false
+}
+
+func getNamespacedResources(
```
yonahd, line 54: Let's rename this function as it no returns whether the namespace is empty

pkg/filters/filters.go
```diff
@@ -107,3 +127,47 @@ func HasIncludedAge(creationTime metav1.Time, filterOpts *Options) (bool, error)
...
+		for _, systemNamespace := range SystemNamespaceNames {
+			if namespaceName == systemNamespace {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func ExcludeNamespacesFilter(object runtime.Object, filterOpts *Options) bool {
```
yonahd, line 144: What's the purpose of the include and exclude namespace functions?

isindir: added comments to both functions

pkg/kor/namespaces.go
isindir: renamed function

isindir:
@yonahd @luisdavim @doronkg, I finally found time and adding some more testing to the namespace removal - but I was not following if json configs with resources to skip from namespace evaluation for different k8s distros is fully implemented or not. Could you please suggest and review if my testing in `pkg/kor/namespacesMoreTests_test.go` is acceptable ? Thanks

yonahd:
> @yonahd @luisdavim @doronkg, I finally found time and adding some more testing to the namespace removal - but I was not following if json configs with resources to skip from namespace evaluation for different k8s distros is fully implemented or not. Could you please suggest and review if my testing in `pkg/kor/namespacesMoreTests_test.go` is acceptable ? Thanks

The resource exceptions are merged fully.
Not sure we will use the namespaces exceptions as we worked around it

yonahd:
Let me know when this is ready for review. This is a massive pr and will take quite a while to review

isindir:
@yonahd I feel that I'll not be able to do 100% unit test coverage for my new code, please review this PR. Thanks.

isindir:
@yonahd - I rebased the PR to the latest, would you have time to review and may be merge it ? We could negotiate next course of actions if needed from my side on a call, please reach me out on keybase , same uid as here.

doronkg:
> > @yonahd - I rebased the PR to the latest, would you have time to review and may be merge it ? We could negotiate next course of actions if needed from my side on a call, please reach me out on keybase , same uid as here.
> 
> 
> 
> Thanks for the PR
> 
> I'm currently very short in time. Any chance @doronkg or @luisdavim can you review this? 

Hi, sorry for the late response.
I'm currently on vacation so my time is very limited. From a quick glimpse - I won't be able to review this PR thoroughly in the following weeks unfortunately.

isindir:
@doronkg would you have some time to review it now ?

doronkg:
> @doronkg would you have some time to review it now ?

Hey, I’m currently traveling, so reviewing everything at once might be a bit challenging. To make this manageable, let’s break it down into smaller chunks and review each component step by step.

So far, I’ve gone through all the additions (except for the tests—we’ll tackle those later) and reviewed the previous comments.

To help me better understand the PR, could you please clarify the following:
	•	Empty namespace evaluation: Could you explain the general logic here so we’re aligned?
	•	Filters logic: How do the new include/exclude/apply filters benefit, since similar logic is already implemented in `namespaces()` under `filters.go`?
	•	Line breaks in function arguments: Are these intentional or a result of your IDE settings?

isindir:
Hi @doronkg , thank you for going through the review, I'll try to answer your questions, and I'm happy to do some rework if needed.

> Empty namespace evaluation: Could you explain the general logic here so we’re aligned?

General logic is defines filter functions and run all of these with the specified parameters for a given namespace from here:
* https://github.com/isindir/kor/blob/unused_namespaces/pkg/kor/namespaces.go#L147 

these functions are:
* by default exclude any system (namespaces created in vanilla Kubernetes); 
* Exclude or Include namespaces
* Filter by `kor` label / any label / age

once we know subjects to inspect - next check is - if namespace contains any non-default resources and resources which user specified to be ignored, which is done here:
* https://github.com/isindir/kor/blob/unused_namespaces/pkg/kor/namespaces.go#L96

and namespaced resource filters are applied here to identify which namespaces can be considered as empty:
* https://github.com/isindir/kor/blob/unused_namespaces/pkg/kor/namespaces.go#L70

> Filters logic: How do the new include/exclude/apply filters benefit, since similar logic is already implemented in namespaces() under filters.go?

I guess this is the function mentioned above:
* https://github.com/isindir/kor/blob/unused_namespaces/pkg/filters/options.go#L114

`Namespaces()` function is called for each for a given resource type to get a list where these resources exist,
for example `Secrets` or `Configmaps`. The function call stack `processNamespaces()` -> `filters.ApplyFilters() , isErrorOrNamespaceContainsResources()` is to get namespaces which are empty (no resources found, not system namespaces and may contain resources to ignore) - list for namespace removal. These 2 functions have slightly different purpose and output may differ based on resources to ignore input parameter as well.

> Line breaks in function arguments: Are these intentional or a result of your IDE settings?

These are not automatic via IDE, but it feels that long strings may read better once refactored:
```diff
-                       fmt.Printf("Do you want to delete %s %s in namespace %s? (Y/N): ", resourceType, resource.Name, namespace)
+                       fmt.Printf(
+                               "Do you want to delete %s %s%s? (Y/N): ",
+                               resourceType,
+                               resource.Name,
+                               namespacedMessageSuffix(namespace),
+                       )
```

> Line breaks in function arguments: Are these intentional or a result of your IDE settings?
> Output examples: Can you show some output examples from your CMD? Relating both flags / edge cases.

I'll try to do this tonight and will send both via new comment.

isindir:
I have a couple of examples now out of the box where it prints unwanted information:
```
% ./build/kor ns --show-reason
kor version: vdev

  _  _____  ____
 | |/ / _ \|  _ \
 | ' / | | | |_) |
 | . \ |_| |  _ <
 |_|\_\___/|_| \_\

Unused resources in namespace: ""
+---+---------------+---------------+------------------+
| # | RESOURCE TYPE | RESOURCE NAME |      REASON      |
+---+---------------+---------------+------------------+
| 1 | Namespaces    | abc           | unused namespace |
+---+---------------+---------------+------------------+
```

and

```
% ./build/kor ns --show-reason -o json
{
  "": {
    "Namespaces": [
      {
        "name": "abc",
        "reason": "unused namespace"
      }
    ]
  }
}
```

isindir:
```
 ./build/kor ns --newer-than 8m
kor version: vdev

  _  _____  ____
 | |/ / _ \|  _ \
 | ' / | | | |_) |
 | . \ |_| |  _ <
 |_|\_\___/|_| \_\

Unused resources in namespace: ""
+---+---------------+---------------+
| # | RESOURCE TYPE | RESOURCE NAME |
+---+---------------+---------------+
| 1 | Namespaces    | abc           |
+---+---------------+---------------+
```

```
% ./build/kor ns --older-than 8m
kor version: vdev

  _  _____  ____
 | |/ / _ \|  _ \
 | ' / | | | |_) |
 | . \ |_| |  _ <
 |_|\_\___/|_| \_\
 ```
 
 ```
 % ./build/kor ns --include-namespaces cde --exclude-namespaces qqq
Exclude namespaces can't be used together with include namespaces. Ignoring --exclude-namespace(-e) flag
kor version: vdev

  _  _____  ____
 | |/ / _ \|  _ \
 | ' / | | | |_) |
 | . \ |_| |  _ <
 |_|\_\___/|_| \_\
 ```
 
 ```
 % ./build/kor ns --delete
Do you want to delete Namespace abc? (Y/N): y
Deleting Namespace abc
kor version: vdev

  _  _____  ____
 | |/ / _ \|  _ \
 | ' / | | | |_) |
 | . \ |_| |  _ <
 |_|\_\___/|_| \_\

Unused resources in namespace: ""
+---+---------------+---------------+
| # | RESOURCE TYPE | RESOURCE NAME |
+---+---------------+---------------+
| 1 | Namespaces    | abc           |
+---+---------------+---------------+
```

doronkg:
Thank you, I have forked your branch to review it more thoroughly and run some tests.
As said, this PR modifies multiple components and I'd like to complete them one by one, so for that matter, we'll break it down to 3 major reviews:
1. Filters
2. Unused Namespaces Evaluation
3. General/Meta

We'll begin with _Filters_, that includes changes mostly in:
- pkg/filters/filters.go
- pkg/filters/filters_test.go
- pkg/kor/namespaces.go

From my impression, all filter-related additions could be reverted or refactored, I'll explain.

### IncludeNamespacesFilter(), ExcludeNamespacesFilter() & ApplyFilters()
As mentioned in https://github.com/yonahd/kor/pull/249#issuecomment-2615164706, this logic is already implemented under `pkg/filters/options.go` by `Namespaces()`, where it's called once and properly sets the desired list of namespaces to evaluate, whether its for the namespaces themselves (i.e. the newly added feature) or other namespaced resources:
```go
// Namespaces returns the namespaces, only called once
func (o *Options) Namespaces(clientset kubernetes.Interface) []string {
	o.once.Do(func() {
		namespaces := make([]string, 0)
		namespacesMap := make(map[string]bool)
		if len(o.IncludeNamespaces) > 0 && len(o.ExcludeNamespaces) > 0 {
			fmt.Fprintf(os.Stderr, "Exclude namespaces can't be used together with include namespaces. Ignoring --exclude-namespaces (-e) flag\n")
			o.ExcludeNamespaces = nil
		}
		includeNamespaces := o.IncludeNamespaces
		excludeNamespaces := o.ExcludeNamespaces
...
```
Therefore, both `IncludeNamespacesFilter()` & `ExcludeNamespacesFilter()` are excessive.
As for the aggregating function - `ApplyFilters()`, there's no need for that function as all the filter flags (i.e. namespaces, age & labels) are covered when passing `filterOpts *filters.Options` to the resource processing.

Let's see an example for that with DaemonSet implementation under `pkg/kor/daemonsets.go`:
```go
func GetUnusedDaemonSets(filterOpts *filters.Options, clientset kubernetes.Interface, outputFormat string, opts common.Opts) (string, error) {
	resources := make(map[string]map[string][]ResourceInfo)
	for _, namespace := range filterOpts.Namespaces(clientset) {
		diff, err := processNamespaceDaemonSets(clientset, namespace, filterOpts)
		if err != nil {
			fmt.Fprintf(os.Stderr, "Failed to process namespace %s: %v\n", namespace, err)
			continue
		}
...
```
In order to process the cluster DaemonSets, we call `range filterOpts.Namespaces(clientset)` to iterate the desired list of namespaces as set by the user (defaults to all unless include/exclude flags are used).
```go
func processNamespaceDaemonSets(clientset kubernetes.Interface, namespace string, filterOpts *filters.Options) ([]ResourceInfo, error) {
	daemonSetsList, err := clientset.AppsV1().DaemonSets(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: filterOpts.IncludeLabels})
	if err != nil {
		return nil, err
	}
	...
        for _, daemonSet := range daemonSetsList.Items {
        		if pass, _ := filter.SetObject(&daemonSet).Run(filterOpts); pass {
	        		continue
		        }
		        if daemonSet.Labels["kor/used"] == "false" {
		        	reason := "Marked with unused label"
			        daemonSetsWithoutReplicas = append(daemonSetsWithoutReplicas, ResourceInfo{Name: daemonSet.Name, Reason: reason})
			        continue
	        	}
...
```
Afterwards, we iterate all the found DaemonSets in each namespace, and apply resource-specific filters such as labels and age with `filter.SetObject(&daemonSet).Run(filterOpts)`.

> __**NOTE:**__ The `KorLabelFilter()` is not applied in this method and still used in most places with a specific condition, we should consolidate it and refactor all occurences, but it's out-of-scope for this PR.

Moving forward, please refactor `pkg/kor/namespaces.go` to that logic and remove these filter functions & their test.

---

### SystemNamespaceFilter()
This filter is of course needed, but its not in the appropriate format, for the following reasons:
1. `filters.go` should be used for generic resource filters, while this is a resource-specific filter.
2. The system namespaces list is hard-coded in the filter instead of using an external resource, which also affects its testing.

The method we use to overcome these requirements is using __Exceptions__, let's return to DaemonSets to better understand it.
Under `pkg/kor/exceptions/` we list all default resources in various K8s distributions to avoid needless scans and false-positive results, in JSON format. For example, `daemonsets/daemonsets.json`:
```json
{
  "exceptionDaemonSets": [
    {
      "Namespace": "kube-system",
      "ResourceName": "aws-node"
    },
...
```
And then, we easily filter these specific resources out in `pkg/kor/daemonsets.go`:
```go
//go:embed exceptions/daemonsets/daemonsets.json
var daemonsetsConfig []byte

func processNamespaceDaemonSets(clientset kubernetes.Interface, namespace string, filterOpts *filters.Options) ([]ResourceInfo, error) {
...
	for _, daemonSet := range daemonSetsList.Items {
		if pass, _ := filter.SetObject(&daemonSet).Run(filterOpts); pass {
			continue
		}

		exceptionFound, err := isResourceException(daemonSet.Name, daemonSet.Namespace, config.ExceptionDaemonSets)
		if err != nil {
			return nil, err
		}

		if exceptionFound {
			continue
		}
...
```
> __**NOTE:**__ Make sure to reference your new JSON [here](https://github.com/yonahd/kor/blob/a0c4f04ccd3dfa0bd5ba75ac12b396c0a8930963/pkg/kor/kor.go#L29) for it to apply.

Same here, please refactor `pkg/kor/namespaces.go` to that logic and remove these filter functions & their tests.

--- 

### In summary
1. Namespace filters should be reverted and used via filter options instead.
2. System namespaces filter should be reverted and used via exceptions instead.
3. Both `filters.go` & `filters_test.go` should remain untouched in this PR.
 
That's it for now, feel free to ask me any questions if something is unclear or if I missed something.
I'm looking forward to review the refactor and move on to reviewing __Unused Namespaces Evaluation__!

Lastly, please update the PR description with the format found under `.github/pull_request_template.md`.

isindir:
@ doronkg , please check now - I think my refactoring for the first batch is what you asked me to do:
* added PR description
* removed my filtering in favour of built-in
* added system namespace filtering via JSON file
* I have not added new tests, but removed all the tests which were made for my filter logic
* I manually tested it, seems to be fine

I've also added one `TODO` mark, where I think refactoring can be done similarly to filtering out namespaces using exception list - but for child resources of the namespace (some hardcoded checks for child resources - https://github.com/yonahd/kor/pull/249/files#diff-ddfa783436a2acc5672cb487e7af28e8a3da84aa480ece960065740409317216R125 ), please let me know if you think this should be refactored using exception lists or similar method. The only concern from my side is that exception list data structure may need some extra information and logic added and can't be used as is.

isindir:
@doronkg , removed - somehow through git manipulations made these changes back and did not double check the final result, sorry

pkg/kor/namespaces.go
```diff
@@ -0,0 +1,246 @@
...
+		fmt.Fprintf(os.Stderr, "Failed to process namespaces: %v\n", err)
+	}
+
+	if len(diff) > 0 {
+		// We consider cluster scope resources in "" (empty string) namespace, as it is common in k8s
+		if resources[""] == nil {
+			resources[""] = make(map[string][]ResourceInfo)
+		}
+		resources[""]["Namespaces"] = diff
+	}
```
doronkg, line 220: This part should be rebuilt by implementing the `GroupBy` option which currently doesn't work:
```
$ kor ns --group-by=resource
kor version: vdev

  _  _____  ____
 | |/ / _ \|  _ \
 | ' / | | | |_) |
 | . \ |_| |  _ <
 |_|\_\___/|_| \_\

Unused s:
+---+------------+---------------+
| # | NAMESPACE  | RESOURCE NAME |
+---+------------+---------------+
| 1 | Namespaces | test          |
+---+------------+---------------+

```
Here's an example from `storageclasses.go`:
```
$ kor sc --group-by=resource
kor version: vdev

  _  _____  ____
 | |/ / _ \|  _ \
 | ' / | | | |_) |
 | . \ |_| |  _ <
 |_|\_\___/|_| \_\

Unused StorageClasss:
+---+-----------+---------------+
| # | NAMESPACE | RESOURCE NAME |
+---+-----------+---------------+
| 1 |           | standard      |
+---+-----------+---------------+
```
```suggestion
	switch opts.GroupBy {
	case "namespace":
		resources[""] = make(map[string][]ResourceInfo)
		resources[""]["Namespace"] = diff
	case "resource":
		appendResources(resources, "Namespace", "", diff)
	}
```

pkg/kor/namespaces_test.go
doronkg, line 1: 1. Fix function tests if needed following `namespaces.go` review.
2. Remove tests for removed functions. 
3. Refactor function naming convention to match the rest of the codebase - `TestFunctionName()` (for example, `Test_namespaces_IgnoreResourceTypes() -> TestIgnoreResourceTypes`).
4. Missing tests for main functions: `processNamespaces()` & `GetUnusedNamespaces()` (see `TestGetUnusedStorageClassesStructured()` for example). Create test namespaces and use `*fake.ClientSet` to evaluate them.

>__**NOTE:**__ Highly recommended to review other resources' tests before working on this.

doronkg (changes requested):
This review covers _Unused Namespaces Evaluation_, I commented on the following files:
- pkg/kor/namespaces.go
- pkg/kor/namespaces_test.go
- pkg/kor/exceptions/namespaces/namespaces.json

As for pkg/kor/namespacesMoreTests_test.go - 
The direction with the different discovery scenarios is very nice and provides solid test coverage, but, it really differs from the method we widely use in our tests, leveraging fake client and creating custom resources for the test suites.
I left no comments here for now, @yonahd I'd like your take on this before moving forward with this test structure.

pkg/kor/exceptions/namespaces/namespaces.json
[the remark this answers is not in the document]
isindir: updated

pkg/kor/namespaces.go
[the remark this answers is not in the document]
isindir: updated

pkg/kor/namespaces.go
```diff
@@ -0,0 +1,246 @@
...
+	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
+
+	"github.com/yonahd/kor/pkg/common"
+	"github.com/yonahd/kor/pkg/filters"
+)
+
+//go:embed exceptions/namespaces/namespaces.json
+var namespacesConfig []byte
+
+type GenericResource struct {
```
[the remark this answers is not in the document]
isindir: updated

pkg/kor/namespaces.go
```diff
@@ -0,0 +1,246 @@
...
+			return unusedNamespaces, err
+		}
+
+		if namespace.Labels["kor/used"] == "false" {
+			unusedNamespaces = append(
+				unusedNamespaces,
+				ResourceInfo{Name: namespace.Name, Reason: "Marked with unused label"},
+			)
+			continue
+		}
```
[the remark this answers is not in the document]
isindir: updated

pkg/kor/namespaces.go
```diff
@@ -0,0 +1,246 @@
...
+			Resource: name,
+		}, nil
+	case 2:
+		return &schema.GroupVersionResource{
+			Group:    splitGV[0],
+			Version:  splitGV[1],
+			Resource: name,
+		}, nil
+	default:
+		return nil, fmt.Errorf("gv is wrong length slice: %d", NumberOfGVPartsFound)
```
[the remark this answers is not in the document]
isindir: updated

pkg/kor/namespaces.go
```diff
@@ -0,0 +1,246 @@
...
+		// User specified resource type ignore list
+		if ignoreResourceType(gr.GVR.Resource, filterOpts.IgnoreResourceTypes) {
+			continue
+		}
+		return true
+	}
+	return false
+}
+
+func isErrorOrNamespaceContainsResources(
```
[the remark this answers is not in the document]
isindir: updated

pkg/kor/namespaces.go
```diff
@@ -0,0 +1,246 @@
...
+		return true
+	}
+	if gr.GVR.Resource == "serviceaccounts" && gr.GVR.Version == "v1" && gr.NamespacedName.Name == "default" {
+		return true
+	}
+	if gr.GVR.Resource == "events" {
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
isindir: updated

yonahd:
> This review covers _Unused Namespaces Evaluation_, I commented on the following files:
> 
> * pkg/kor/namespaces.go
> * pkg/kor/namespaces_test.go
> * pkg/kor/exceptions/namespaces/namespaces.json
> 
> As for pkg/kor/namespacesMoreTests_test.go - The direction with the different discovery scenarios is very nice and provides solid test coverage, but, it really differs from the method we widely use in our tests, leveraging fake client and creating custom resources for the test suites. I left no comments here for now, @yonahd I'd like your take on this before moving forward with this test structure.



> This review covers _Unused Namespaces Evaluation_, I commented on the following files:
> 
> * pkg/kor/namespaces.go
> * pkg/kor/namespaces_test.go
> * pkg/kor/exceptions/namespaces/namespaces.json
> 
> As for pkg/kor/namespacesMoreTests_test.go - The direction with the different discovery scenarios is very nice and provides solid test coverage, but, it really differs from the method we widely use in our tests, leveraging fake client and creating custom resources for the test suites. I left no comments here for now, @yonahd I'd like your take on this before moving forward with this test structure.

The structure of our tests reflects high coverage UT
We want to hit all the edge cases in the code in the most simple way
I'd say we can simplify the tests to create simple scenarios like we do in the rest of the project touching on all the edge cases.
Focusing on each namespace with resources 
Suggestion for tests
**isNamespaceNotEmpty function**
- Namespace contains non-ignored resource.
- Namespace contains only ignored resources.
- Namespace contains event resource (which should be ignored).
- Namespace is empty (no resources).

**isNamespaceUsed function**
- Namespace contains used resources.
- Namespace contains only ignored resources.
- Namespace is empty (no resources).
- Error during listing resources (should skip that resource).
- Multiple resource types to check.
- Empty namespace with many resource types (edge case).

yonahd:
> This review covers _Unused Namespaces Evaluation_, I commented on the following files:
> 
> * pkg/kor/namespaces.go
> * pkg/kor/namespaces_test.go
> * pkg/kor/exceptions/namespaces/namespaces.json
> 
> As for pkg/kor/namespacesMoreTests_test.go - The direction with the different discovery scenarios is very nice and provides solid test coverage, but, it really differs from the method we widely use in our tests, leveraging fake client and creating custom resources for the test suites. I left no comments here for now, @yonahd I'd like your take on this before moving forward with this test structure.

The structure of our tests reflects high coverage UT
We want to hit all the edge cases in the code in the most simple way
I'd say we can simplify the tests to create simple scenarios like we do in the rest of the project touching on all the edge cases.
Focusing on each namespace with resources
Suggestion for tests:
**isNamespaceNotEmpty function**
- Namespace contains non-ignored resource.
- Namespace contains only ignored resources.
- Namespace contains event resource (which should be ignored).
- Namespace is empty (no resources).

**isNamespaceUsed function**
- Namespace contains used resources.
- Namespace contains only ignored resources.
- Namespace is empty (no resources).
- Error during listing resources (should skip that resource).
- Multiple resource types to check.
- Empty namespace with many resource types (edge case).

pkg/kor/namespaces.go
```diff
@@ -0,0 +1,246 @@
...
+	apiResourceLists, err := clientset.Discovery().ServerPreferredNamespacedResources()
+	if err != nil {
+		return true, err
+	}
+
+	// Iterate over all API resources and list instances of each in the specified namespace
+	for _, apiResourceList := range apiResourceLists {
+		for _, apiResource := range apiResourceList.APIResources {
+			gv := strings.Split(apiResourceList.GroupVersion, "/")
+			gvr, err := getGVR(apiResource.Name, gv)
```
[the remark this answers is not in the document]
isindir: updated

pkg/kor/namespaces.go
```diff
@@ -0,0 +1,246 @@
...
+			if err != nil {
+				return true, err
+			}
+
+			unstructuredList, err := dynamicClient.Resource(*gvr).Namespace(namespace).List(ctx, metav1.ListOptions{})
```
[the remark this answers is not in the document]
isindir: updated

pkg/kor/namespaces_test.go
isindir: 1. done along the process
2. done along the process

pkg/kor/namespaces.go
```diff
@@ -0,0 +1,246 @@
...
+	if gr.GVR.Resource == "serviceaccounts" && gr.GVR.Version == "v1" && gr.NamespacedName.Name == "default" {
+		return true
+	}
+	if gr.GVR.Resource == "events" {
+		return true
+	}
+	return false
+}
+
+func isNamespaceNotEmpty(
```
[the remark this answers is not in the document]
isindir: updated

pkg/kor/namespaces_test.go
isindir: 3. updated

isindir:
@yonahd and @doronkg ,

Thank you for reviews! I'm still working on it :)

> Suggestion for tests: 
> **isNamespaceNotEmpty function**
> **isNamespaceUsed function**

@yonahd , As I was requested on a previous review by @doronkg to merge both functions, now, after merging, I need to cover single function with all mentioned coverage tests.

isindir:
Hello, @yonahd and @doronkg ,
I hit some edge case in `kor` and need advice here. General logic for namespaces is as follows:
* get all namespaces and iterate over each
  * filter out namespaces by user and kor specified criteria
  * for each non-filtered namespace get all resources of all types
    * !!! here in my original code I was applying filters myself including "by default" implied filters, but now I can only do filtering by user defined filters, as any custom logic is gone now. I'd like to filter using json exception lists, can it be implemented relatively simply, or I need to introduce new piece of custom logic to use json exception lists, wdyt ?

here is location where I think it is needed: https://github.com/yonahd/kor/pull/249/files#diff-ddfa783436a2acc5672cb487e7af28e8a3da84aa480ece960065740409317216R151

I'll try to figure it out, but it seems for me - namespaces are quite a unique resource in this case, as these are hosting resources rtaher then just have child resources - and I will need to introduce custom logic to utilise all exception lists (use existing exception list for a specific found resource).

doronkg:
If I understand you correctly, you refer to default resources found in each namespace in different k8s distros (i.e. default ServiceAccount, etc) in addition to the default namespaces themselves (i.e default etc).

In that case, we can use the current exception lists. Default namespaces are already ignored, as for resources that should be ignored from all namespaces - they are specified with RegEx as so:

```

    {
      "Namespace": ".*",
      "ResourceName": "example"
      "MatchRegex": true
    }
```

It could be better for maintenance to somehow aggregate all these instances and form a (kind, name) collection which will be matched against in the filtering, instead of an additional hard coded list in the code.

For example, it would concise [(ServiceAccount, default), (ConfigMap, openshift-service-ca.crt)].

With that being said, this list is very short and might not be worthy of the addition effort for now, and it might be easier to go with the current implementation and leave it for the future.

@yonahd wdyt?

isindir:
I also think it is better to finalize tests for the current PR and refactor Config and add dynamic list generation as a separate PR. This PR is a big change on it's own.

doronkg:
> I also think it is better to finalize tests for the current PR and refactor Config and add dynamic list generation as a separate PR. This PR is a big change on it's own.

So for now reference in your code all the resources that match this pattern from the exception lists:
```json
    {
      "Namespace": ".*",
      "MatchRegex": true
    }
```

isindir:
> So for now reference in your code all the resources that match this pattern from the exception lists:
> 
> ```json
>     {
>       "Namespace": ".*",
>       "MatchRegex": true
>     }
> ```

@doronkg - implemented, please review. I had to create a new structure and create a "copy" of the function which is performing regex match.

doronkg:
ACK, thanks, I'll try to review it ASAP.

pkg/kor/exceptions/namespaced-resources/namespaced-resources.json
yonahd, line 1: Shouldn't we just use the exceptions files for all resources?

isindir: These files themselves do not contain resource type, some are for namespaced resources and some other for cluster level resources - for me, it feels natural to have separate function for filtering and separate list of resources for fairly unique use case. Otherwise, all other json files need refactoring.

yonahd: Why do they need refactoring?
Any resource from any of the exception files found in a namespace would not make the namespace considered used

isindir: I think it does not make sense to go over each file and perform a comparison check on each exclusion, it is very small subset of the checks which are needed for identifying default namespaced resources - or resources which can be safely ignored like events. Also, not all the file names following same convention as k8s resource names, for example pdb, so without touching it - it would be impossible to use existing embedded blobs anyway. If you still want me to change the behaviour - please let me know how, and I will refactor it.

yonahd: I understand where you are coming from but the current implementation has duplication and high potential to miss things
e.g. This might be something we missed https://github.com/yonahd/kor/blob/main/pkg/kor/exceptions/roles/roles.json#L4

What do you suggest we use as a safeguard?

isindir: I feel there is a need for a flag to say the set of objects is namespaced like `kc api-resources` command sets that field - so these could be ignored out of the box. Then I think there is potential in go generate to prefetch the list of wildcarded resources we are interested only in this specific case, taken in consideration we know which resources are namespaced (we could of course ask k8s API to tell us about each of these, but still need to converge file/variable names to be 100% comptabile with k8s to use reflect for example). btw, if you have keybase, in my github profile I have keybase handle reference, we can discuss offline how to better resolve this.

Issue #253, opened 2024-05-01
Refactor: move resource exceptions to external files 


yonahd:
Move the resource exceptions to a different file. 

Currently the lists of exception resource are quite long e.g. https://github.com/yonahd/kor/blob/main/pkg/kor/configmaps.go#L17 and will cause future contributors more cognitive load.

We should separate these from the actual code

pbr0ck3r:
It might be nice to also break out resources into their individual clouds (then merge them all together), so we can better track changes related to specific environments.

- pkg
  - exceptions
    - gke
    - aks
    - kind
    - minikube
    - universal (i.e. kube-proxy)

yonahd:
> It might be nice to also break out resources into their individual clouds (then merge them all together) using a helper function. We can better track changes related to specific environments this way and also allow targeting specific environments.
> 
> * pkg
>   
>   * exceptions
>     
>     * gke
>     * aks
>     * kind
>     * minikube
>     * universal (i.e. kube-proxy)

I was thinking of splitting it using the resource type but this might make more sense. Let's see if there are any takers and we'll see how simply it can be done

pbr0ck3r:
Happy to take a initial stab at this.

yonahd:
Go for it.
Let's see how it looks

PR #257, opened 2024-05-02
fix: refactor exceptions into seperate files

pbr0ck3r:
Please read [CONTRIBUTING.md](CONTRIBUTING.md) for additional information on contributing to this repository!

- initial take on how we could refactor exceptions into their own files using JSON.

## PR Checklist

- [x] This PR adds new code

## Github Issue

https://github.com/yonahd/kor/issues/253


pkg/kor/kor.go
```diff
@@ -27,6 +28,11 @@ type IncludeExcludeLists struct {
 	ExcludeListStr string
 }
 
+type Config struct {
```
yonahd, line 31: Interesting. Did you have something in mind to support using this struct?

pkg/kor/exceptions/storageclasses/config.json
```diff
@@ -0,0 +1,16 @@
+{
```
yonahd, line 1: Nit: maybe storageclasses.json?

yonahd (commented):
Looks simple and straightforward.
A couple points here:

- Maybe we can add a field of source which would be more for documentation e.g. GKE or minikube
- The folder name we call it exceptions, trying to think if there is a better name for the false positives

pkg/kor/exceptions/storageclasses/config.json
pbr0ck3r: Happy to make this change!

pkg/kor/kor.go
pbr0ck3r: It is used by `unmarshalConfig`. But happy to refactor that into not using it. The idea is that this sets us up in a way we could break out exceptions into individual cloud if we want in the future. Then we could adjust the code to do say GKE only (if you pass in a flag) or all by default.

pbr0ck3r: If we don't use a struct like this, then we will have to have each embedded file be its own struct in each of their individual files to unmarshal the data.

```
type Config struct {
	ExceptionStorageClasses []ExceptionResource `json:"exceptionStorageClasses"`
}

type ExceptionClusterRoles struct {
	ExceptionClusterRoles []ExceptionResource `json:"exceptionClusterRoles"`
}
```

Which will then require us to refactor the `unmarshalConfig` helper.

The other idea would be do away with JSON files altogether and just  create another Go package called `exceptions` and we move everything, i.e. `pkg/exceptions/exceptionStorageClasses.go`, `pkg/exceptions/exceptionClusterRoles.go`

yonahd: > It is used by `unmarshalConfig`. But happy to refactor that into not using it. The idea is that this sets us up in a way we could break out exceptions into individual cloud if we want in the future. Then we could adjust the code to do say GKE only (if you pass in a flag) or all by default.

I actually did not think of that. Really solid idea!

pbr0ck3r: I pushed up some more code to show how the `Config` struct will look in the future.

yonahd: Looks good to me. Go for it

yonahd (approved):
Lgtm
Thank you!

PR #262, opened 2024-05-06
feat: add OpenShift exceptions

doronkg:
## What this PR does / why we need it
This PR excludes the default resources created in basic OpenShift installations.
It also includes a fix in `cmd/kor/crds.go` to allow a new plural alias - `kor crds`.

## PR Checklist

- [x] This PR adds K8s exceptions (false positives)
- [ ] This PR adds new code
- [ ] This PR includes test for any new code

## Github Issue

Closes #240

## Notes for your reviewers
- In `pkg/kor/secrets.go`, a new exception secret type was added - `kubernetes.io/dockercfg`, which is the OpenShift equivalent of `kubernetes.io/dockerconfigjson`.

- This PR addresses all default namespaces that **doesn't** begin with `openshift-` prefix.
Meaning, it covers the following namespaces: `openshift`, `default`, `kube-system`, `kube-public`, `kube-node-lease` & `assisted-installer`.

Basic OpenShift installation comes with 60+ namespaces beginning with `openshift-` prefix, which doesn't include additional namespaces created by OpenShift operators or customized installations, that would also be created with that prefix.

As I see it, there are 3 options to address this case:
1. Entirely exclude all namespaces with `openshift-` prefix (flexible).
2. Entirely exclude all the default namespaces created in the basic installation.
3. Exclude all the default resources created in the default namespaces.

@yonahd please share your thoughts, I'm leaning towards option no. (1).

PR #270, opened 2024-05-18
feat: add K3S exceptions

### LINKED ISSUE #259: Map false unused resources: K3s
Find all default resources in a K3s cluster that kor flags as unused.
List them
(example can be found here https://github.com/yonahd/kor/blob/main/pkg/kor/serviceaccounts.go#L16)

Itaykal:
Adds exceptions to all k3s default objects.

## PR Checklist

- [x] This PR adds K8s exceptions (false positives)
- [ ] This PR adds new code

## Github Issue

Fixes #259

## Notes for your reviewers
Theres a default secret which is prefixed by the node's name. (e.g: `nodename.node-password.k3s`)
As wildcards work only as an indicator for a suffix I can't really really ignore it at the current state. <br />
I think the way forward will be one of the following options:

1. Allow wildcards/regex in the exception handling.
1. Allow `*` to be used as a prefix as well as a suffix.

I think option 1 has more risk of ignoring objects that aren't supposed to be ignored, so the second one seems more appealing to me. 
On the other hand, the second option is not as straightforward as well-known and used wildcards.

### DIFF
--- a/pkg/kor/exceptions/clusterroles/clusterroles.json
+++ b/pkg/kor/exceptions/clusterroles/clusterroles.json
@@ -83,6 +83,14 @@
     {
       "Namespace": "",
       "ResourceName": "view"
+    },
+    {
+      "Namespace": "",
+      "ResourceName": "system:node-bootstrapper"
+    },
+    {
+      "Namespace": "",
+      "ResourceName": "system:aggregated-metrics-reader"
     }
   ]
 }
--- a/pkg/kor/exceptions/configmaps/configmaps.json
+++ b/pkg/kor/exceptions/configmaps/configmaps.json
@@ -79,6 +79,10 @@
     {
       "Namespace": "kubernetes-dashboard",
       "ResourceName": "kubernetes-dashboard-settings"
+    },
+    {
+      "Namespace": "kube-system",
+      "ResourceName": "cluster-dns"
     }
   ]
 }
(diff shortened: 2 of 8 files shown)


yonahd:
Pull request looks good.
To your point regarding nodename.node-password.k3s
We were able to avoid regexes so far using the secret type(e.g. helm)
Is there any chance these have a different type than the default opaque?

Otherwise we will need to consider regexes

Itaykal:
> Is there any chance these have a different type than the default opaque?

Unfortunately it's opaque.

yonahd:
@doronkg we were looking at a similar issue with the openshift namespaces right?

doronkg:
> @doronkg we were looking at a similar issue with the openshift namespaces right?

Not sure which case you meant so I'll address both:

As for regex:
>We were able to avoid regexes so far using the secret type(e.g. helm)

In OpenShift, we used type `kubernetes.io/dockercfg` to avoid this similar case.

In addition, as described in #262:
>Basic OpenShift installation comes with 60+ namespaces beginning with `openshift-` prefix, which doesn't include additional namespaces created by OpenShift operators or customized installations, that would also be created with that prefix.

In this case we discussed excluding all the `openshift-` prefix namespaces, as they are loaded with default unused resources.

Itaykal:
Well, the object resides in `kube-system` so ignoring it by namespace won't be possible, and as mentioned it is of type `opaque` so both solutions are not relevant for this case.

yonahd:
Added a ticket for this.
We can either merge this now and then we can put the information for the relevant resources on that ticket. Or we can wait

yonahd:
OK. Just add this secret regex in #277 and use it as a test case

yonahd (approved):

PR #277, opened 2024-05-20
feat: add regex in resource exceptions

### LINKED ISSUE #276: Feature: support regexes in resource exceptions
**Is your feature request related to a problem? Please describe.**
When mapping the false positives in different kubernetes distributions we found some resources we couldn't exclude due to the name being dependent on random or specific (e.g. Node name) 
Support excluding these resources 
.

Itaykal:
Adds regular expression support to `exceptions/*/*.json` files.

## PR Checklist

- [X] This PR adds new code
- [x] This PR includes test for any new code

## Github Issue
Fixes #276 

## Notes for your reviewers
Not sure this is how this impacts performance.
The behavior of the exception filtration in configmaps and serviceaccounts is now changed to be aligned with all other objects, this as well has some impact on performance.


yonahd:
The PR looks like it would work well.
The issue I have with it is the readability, searchability and the extra work to add new resources here.
The question is if supporting prefixes and suffixes would be better codewise as that would cover the functionality we are currently lacking

Itaykal:
It's quite frustrating to work with...
I thought about adding a flag to enable regex searching on specific exceptions that require them and thus simplifying all "non-regex" searches, something along the lines of this:
```json
{
    "Namespace": ".*-example",
    "ResourceName": "example",
    "Regex": true
}
```
This is also quite unreadable but it grants the efficiency and ease of use of regex while still allowing more readable exceptions when not needed.

Another topic that scares me a bit more then readability is that with some minor human error this could cause false positives.
I think this should be discussed before moving on to readability (and ease of use) as I don't see any way to avoid this risk.

yonahd:
> It's quite frustrating to work with... I thought about adding a flag to enable regex searching on specific exceptions that require them and thus simplifying all "non-regex" searches, something along the lines of this:
> 
> ```json
> {
>     "Namespace": ".*-example",
>     "ResourceName": "example",
>     "Regex": true
> }
> ```
> 
> This is also quite unreadable but it grants the efficiency and ease of use of regex while still allowing more readable exceptions when not needed.
> 
> Another topic that scares me a bit more then readability is that with some minor human error this could cause false positives. I think this should be discussed before moving on to readability (and ease of use) as I don't see any way to avoid this risk.

The damage of a regex too broad would marking unused resources as used which is not great but the damage is less than the reverse.

We obviously want kor to find as many unused resources as possible so we would need to be cautious in this area.

The solution might be to create a separate json for regexes or prefixes and handle it separately

Itaykal:
I've implemented the flag solution as a flag as it's easier to handle later on, wdyt?

yonahd:
Looks much better.
I'd like to have tests in place for this.
Also (not necessarily belonging to this PR but we need to have validators for the exception jsons)

doronkg:
@Itaykal can you rebase the branch so the exceptions folder will go under validation of the new GitHub action added in #295?

yonahd:
> What should be tested here? the regex exception's usability or the validity of the expressions themselves?

I think we need to test the that it excludes 
- regular resource
- regex resource name with regular namespace
- regex namespace regular resource name

kor
yonahd, line 1: This needs to be removed

pkg/kor/kor_test.go
```diff
@@ -130,3 +130,55 @@ func TestGetKubeClientFromInput(t *testing.T) {
 		t.Errorf("Expected valid clientSet")
 	}
 }
+
```
yonahd, line 133: Why not create resource specific tests?

yonahd (commented):
Couple of comments

pkg/kor/kor_test.go
Itaykal: What do you mean by resource specific?

yonahd: For example create a configmap with the name /regex from the exceptions json

Itaykal: if I'm understanding correctly, you're suggesting to test with the objects themselves and not just the name,

If that's truly the case, I'm not sure I see a point in doing so. The functionality of resource exceptions should be the same over all objects and thus only it should be tested and not the way the parameters to said procedure are provided.

Apologies for my delayed response,

yonahd: OK.
Makes sense.
I'll make some order in this kor file soon and split it off, but logically this does cover your code

yonahd (approved):
Lgtm
Thanks!

doronkg:
@Itaykal many exception duplications were added in this PR (probably following the rebasing), can you create a new PR that removes them?

Bonus - enhance the validation workflow (#298) to remove duplications.

PR #283, opened 2024-05-22
Feat: add failed jobs as unused

yonahd:
This PR adds a case where A Job has failed and reached it's maximum retries.
This pod should now be considered unused

## PR Checklist

- [x] This PR adds new code
- [ ] This PR adds docs


doronkg:
PR looks solid, few thoughts, as we're shifting the way we approach unused Jobs from just completed ones:
1. If adding failures as an indication for unused, let's add another condition for `DeadlineExceeded`.
2. We can also consider checking [FailedIndexes](https://kubernetes.io/docs/concepts/workloads/controllers/job/#backoff-limit-per-index) which is another type of BackoffLimit.
3. What about suspended Jobs?

pkg/kor/jobs.go
```diff
@@ -42,6 +43,19 @@ func processNamespaceJobs(clientset kubernetes.Interface, namespace string, filt
...
+			backoffExceeded := false
+			for _, condition := range job.Status.Conditions {
+				if condition.Type == batchv1.JobFailed && condition.Reason == "BackoffLimitExceeded" {
+					backoffExceeded = true
+					break
+				}
+			}
+			if backoffExceeded {
+				unusedJobNames = append(unusedJobNames, job.Name)
+			}
```
doronkg, line 58:
```suggestion
			for _, condition := range job.Status.Conditions {
				if condition.Type == batchv1.JobFailed && condition.Reason == "BackoffLimitExceeded" {
					unusedJobNames = append(unusedJobNames, job.Name)
					break
				}
			}
```

pkg/kor/jobs_test.go
```diff
@@ -71,23 +71,44 @@ func createTestJobs(t *testing.T) *fake.Clientset {
...
+	if len(unusedJobs) != 3 {
+		t.Errorf("Expected 3 job been completed, got %d", len(unusedJobs))
 	}
 
-	if completedJobs[0] != "test-job2" && completedJobs[1] != "test-job4" {
-		t.Errorf("job2', got %s", completedJobs[0])
+	expectedJobs := []string{"test-job2", "test-job4", "test-job5"}
+	for _, jobName := range expectedJobs {
+		if !contains(unusedJobs, jobName) {
+			t.Errorf("Expected job %s to be completed, but it was not found", jobName)
```
doronkg, line 110: The error messages should indicate the new failure condition as well, not only completion.

pkg/kor/jobs.go
yonahd: Good catch. Tired coding

pkg/kor/jobs_test.go
```diff
@@ -71,23 +71,44 @@ func createTestJobs(t *testing.T) *fake.Clientset {
...
-	completedJobs, err := processNamespaceJobs(clientset, testNamespace, &filters.Options{})
+	unusedJobs, err := processNamespaceJobs(clientset, testNamespace, &filters.Options{})
 	if err != nil {
 		t.Errorf("Expected no error, got %v", err)
 	}
 
-	if len(completedJobs) != 2 {
-		t.Errorf("Expected 2 job been completed, got %d", len(completedJobs))
+	if len(unusedJobs) != 3 {
+		t.Errorf("Expected 3 job been completed, got %d", len(unusedJobs))
```
doronkg, line 104: Also this error message :)

Issue #286, opened 2024-05-23
Test: Add test for Helm chart


doronkg:
Currently, the kor Helm chart is not covered with tests.
When modifying resources, there's no automatic validation that the chart could be successfully rendered / deployed.

Add chart tests e.g. `helm template` / gh-actions.

found-it:
@yonahd I took a crack at this one with https://github.com/yonahd/kor/pull/293. Happy to party on that PR more if you'd like!

PR #292, opened 2024-05-29
fix(helm): Add missing labels key in metadata

idfer:

Fix missing `labels` key on cronjob.yaml that cause YAML parse error when run `helm` command.

## PR Checklist

- [ ] This PR adds new code
- [ ] This PR adds docs
- [x] This PR fix helm chart template error

### DIFF
--- a/charts/kor/templates/cronjob.yaml
+++ b/charts/kor/templates/cronjob.yaml
@@ -12,7 +12,8 @@ spec:
     spec:
       template:
         metadata:
-          {{- include "kor.labels" . | nindent 8 }}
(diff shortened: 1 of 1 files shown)


yonahd:
Thanks!
Would you like to take on this https://github.com/yonahd/kor/issues/286
So things like this don't reaccur
It can be in a separate pr

found-it:
@yonahd any chance we can merge this PR? I added some helm chart testing and if all goes to plan, it'll block until these fixes are in

yonahd (approved):

Issue #287, opened 2024-05-23
Chore: Run helm-docs as GitHub action & bump versions


doronkg:
Currently, when contributors make alternations in kor Helm chart values, the chart's README should be updated manually as part of the PR.

Add `helm-docs` as a GitHub action to allow seamlessly docs updates, either by using a community gh-action or a custom workflow.

In addition, bump both `version` & `appVersion`.

References:
[#280](https://github.com/yonahd/kor/pull/280)
https://github.com/shaybentk/helm-docs-action

PR #293, opened 2024-05-29
feat: add helm chart testing

found-it:
Add helm chart testing with https://github.com/helm/chart-testing-action

This PR adds some basic helm chart testing. It lints the chart and spins up a cluster to install the chart with each values file in `charts/kor/ci/*-values.yaml`.

## PR Checklist

- [ ] This PR adds new code
- [ ] This PR adds docs
- [x] This PR add helm chart tests

## GitHub Issue

Should address https://github.com/yonahd/kor/issues/286

## Notes for your reviewers

I just took the basic example from https://github.com/helm/chart-testing-action and applied it here. I ran `ct install` locally to test the chart, it correctly failed without the fixes from https://github.com/yonahd/kor/pull/292 and correctly passed when I applied the fixes.

I just set up default and "with cronjob" values since those are the two documented cases in the README.

I also had to modify the `.gitignore` because the `kor` entry was too broad and ended up ignoring things in `charts/kor`

### DIFF
--- a/.github/workflows/helm-test.yaml
+++ b/.github/workflows/helm-test.yaml
@@ -28,7 +28,7 @@ jobs:
           check-latest: true
 
       - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.7.0
+        uses: helm/chart-testing-action@v2.6.1
 
       - name: Run chart-testing (list-changed)
         id: list-changed
--- a/.github/workflows/helm-test.yaml
+++ b/.github/workflows/helm-test.yaml
@@ -46,7 +46,7 @@ jobs:
         if: steps.list-changed.outputs.changed == 'true'
         run: |
           docker run --rm --volume "$(pwd):/helm-docs" jnorwood/helm-docs:latest
-          if ! git diff-index --quiet HEAD --; then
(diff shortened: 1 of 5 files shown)


yonahd:
Looks like you need to bump the chart version
And then regenerate the Readme using
`docker run --rm --volume "$(pwd):/helm-docs" -u $(id -u) jnorwood/helm-docs:latest `

yonahd:
Looks good.
Just regenerate the helm Readme using
docker run --rm --volume "$(pwd):/helm-docs" -u $(id -u) jnorwood/helm-docs:latest

found-it:
@yonahd I added a test to CI to alert when the readme needs to be regenerated. I just want to test and make sure it does the right thing then I will update the readme and add

found-it:
Awesome, looks like it did the right thing. Updated the readme

yonahd:
> @yonahd I added a test to CI to alert when the readme needs to be regenerated. I just want to test and make sure it does the right thing then I will update the readme and add

Nice!
Looks like you also did most of #287

yonahd (approved):
LGTM
Thank you!

PR #311, opened 2024-06-18
feat: Add OpenShift exceptions

### LINKED ISSUE #240: Map false unused resources: Openshift
Find all default resources in a Openshift cluster that kor flags as unused.
List them
(example can be found here  https://github.com/yonahd/kor/blob/main/pkg/kor/serviceaccounts.go#L16)

doronkg:
This PR excludes the default resources created in basic OpenShift installations.
In addition, I've moved the config load out of the processing loop in most resource types, to reduce excessive iterations.

## PR Checklist

- [x] This PR adds K8s exceptions (false positives)
- [x] This PR adds new code

## GitHub Issue

Closes #240

## Notes for your reviewers
In `pkg/kor/secrets.go`, a new exception secret type was added - `kubernetes.io/dockercfg`, which is the OpenShift equivalent of `kubernetes.io/dockerconfigjson`.

Basic OpenShift installation comes with 60+ namespaces beginning with `openshift-` prefix, which doesn't include additional namespaces created by OpenShift operators or customized installations, that would also be created with that prefix.
For that case, I've excluded all `openshift-` namespaces in all relevant namespaced resource types, with the following pattern:
```
    {
      "Namespace": "openshift-.*",
      "ResourceName": ".*",
      "MatchRegex": true
    }
```

### DIFF
--- a/pkg/kor/exceptions/pdbs/pdbs.json
+++ b/pkg/kor/exceptions/pdbs/pdbs.json
@@ -1,0 +1,9 @@
+{
+  "exceptionPdbs": [
+    {
+      "Namespace": "openshift-.*",
+      "ResourceName": ".*",
+      "MatchRegex": true
+    }
+  ]
+}
--- a/pkg/kor/exceptions/replicasets/replicasets.json
+++ b/pkg/kor/exceptions/replicasets/replicasets.json
@@ -1,0 +1,9 @@
+{
+  "exceptionReplicaSets": [
+    {
+      "Namespace": "openshift-.*",
+      "ResourceName": ".*",
+      "MatchRegex": true
+    }
+  ]
+}
--- a/pkg/kor/exceptions/replicasets/replicasets.json
+++ b/pkg/kor/exceptions/replicasets/replicasets.json
@@ -1,9 +1,0 @@
-{
-  "exceptionReplicaSets": [
-    {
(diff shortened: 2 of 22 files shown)


pkg/kor/exceptions/replicasets/replicasets.json
```diff
@@ -0,0 +1,9 @@
+{
+  "exceptionReplicaSets": [
```
yonahd, line 2: Should we be keeping empty replicasets?

pkg/kor/exceptions/pdbs/pdbs.json
```diff
@@ -0,0 +1,9 @@
+{
+  "exceptionPdbs": [
+    {
+      "Namespace": "openshift-.*",
```
yonahd, line 4: Can you add a test for both in regex?
We don't have a test for this case

yonahd (commented):
Couple small comments

pkg/kor/exceptions/replicasets/replicasets.json
doronkg: When I come to think about it, except for optional rollback (less common with default installation ones), there's no much added value in preserving unused ReplicaSets. I'll revert this addition, encouraging users to remove them and free up etcd space.

yonahd (approved):
Lgtm

PR #360, opened 2024-09-17
When user can't list return empty

yonahd:
When user can't list namespaces return empty instead of exiting

## PR Checklist

- [x] This PR adds new code
- [ ] This PR adds docs


yonahd:
@hagay3 Is this good enough?

PR #300, opened 2024-06-14
refactor(cmd): standardize subcommand aliases with kubectl shortnames

doronkg:
## What this PR does / why we need it?
This PR refactors subcommand aliases to match kubectl official shortnames, as disaplyed in `kubectl api-resources`.
Fixes:
- Remove excess singular aliases (i.e. "job", "replicaset")
- Remove unofficial shortnames (i.e. "scrt")
- Support uncovered shortnames (i.e. "crds")

In addition, this PR refactors the _Usage_ section in `README.md` to address kor's subcommands in singular form.

## PR Checklist

- [x] This PR adds new code
- [x] This PR adds docs

## GitHub Issue

[XX-XX]

## Notes for your reviewers
<details><summary>Kor Original Aliases</summary>

```console
$ grep -r Aliases cmd/kor/ -B 1
cmd/kor/clusterroles.go-        Use:     "clusterrole",
cmd/kor/clusterroles.go:        Aliases: []string{"clusterroles"},
--
cmd/kor/configmaps.go-  Use:     "configmap",
cmd/kor/configmaps.go:  Aliases: []string{"cm", "configmaps"},
--
cmd/kor/crds.go-        Use:     "customresourcedefinition",
cmd/kor/crds.go:        Aliases: []string{"crd", "customresourcedefinitions"},
--
cmd/kor/daemonsets.go-  Use:     "daemonset",
cmd/kor/daemonsets.go:  Aliases: []string{"ds", "daemonsets"},
--
cmd/kor/deployments.go- Use:     "deployment",
cmd/kor/deployments.go: Aliases: []string{"deploy", "deployments"},
--
cmd/kor/finalizers.go-  Use:     "finalizer",
cmd/kor/finalizers.go:  Aliases: []string{"fin", "finalizers"},
--
cmd/kor/hpas.go-        Use:     "horizontalpodautoscaler",
cmd/kor/hpas.go:        Aliases: []string{"hpa", "horizontalpodautoscalers"},
--
cmd/kor/ingresses.go-   Use:     "ingress",
cmd/kor/ingresses.go:   Aliases: []string{"ing", "ingresses"},
--
cmd/kor/jobs.go-        Use:     "job",
cmd/kor/jobs.go:        Aliases: []string{"job", "jobs"},
--
cmd/kor/pdbs.go-        Use:     "poddisruptionbudget",
cmd/kor/pdbs.go:        Aliases: []string{"pdb", "poddisruptionbudgets"},
--
[...]

PR #362, opened 2024-09-30
Feat: Discover unused RoleBindings

### LINKED ISSUE #334: feature: Discover unused RoleBindings
**Is your feature request related to a problem?**
RoleBindings can be created while referencing **non-existing** users/groups/ServiceAccounts on one hand or Roles/ClusterRoles on the other hand.

Go through all existing RoleBindings and verify if they are applied to existing subjects and roles.
Utilize `ShowReason` flag to indicate that the reason the RoleBinding was considered unused was because it referenced an unused subject or role.

__**NOTE:**__ Since a RoleBinding can include multiple subject references, discovering a single non-existing subject (one of several existing ones) might indicate the RoleBinding as unused while it actually is, in that case, it shouldn't be considered as unused.

<details>
<summary>Examples</summary>

In the attached example, we could see a RoleBinding with references to both users: `alice` & `bob`.
1. Assuming both users `alice` & `bob` does not exist, the Role is not applied to them, hence the RoleBinding will be considered as **UNUSED**.
2. Assuming both users `alice` & `bob` exist, but the Role does not exist, it is not applied to them, hence the RoleBinding will be considered as **UNUSED**.
3. Assuming user `alice` does exist and the Role is applied to it, even while `bob` does not exist - the RoleBinding will be considered as **USED**.

```
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: read-pods
  namespace: default
subjects:
- kind: User
  name: alice
  apiGroup: rbac.authorization.k8s.io
- kind: User
  name: bob
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: Role 
  name: pod-reader 
  apiGroup: rbac.authorization.k8s.io
```

</details>

**Describe the solution you'd like**
```console
$ kor rolebindings
Unused RoleBindings:
+---+----------------+----------------------------------------------+-----------------------------+
| # | NAMESPACE      |   RESOURCE NAME    |                         REASON                        |
+---+----------------+----------------------------------------------+-----------------------------+
| 1 | example-ns-1   | example-rb-1       | RoleBinding references a non-existing ServiceAccount  |  
| 2 | example-ns-2   | example-rb-2       | RoleBinding references a non-existing ClusterRole     |
+---+----------------+----------------------------------------------+-----------------------------+
```
**Feature checklist**
- [ ] `pkg/kor/rolebindings.go`
- [ ] `pkg/kor/rolebindings_test.go`
- [ ] `pkg/kor/create_test_resources.go`
- [ ] `pkg/kor/all.go`
- [ ] `pkg/kor/delete.go`
- [ ] `pkg/kor/multi.go`
- [ ] `cmd/kor/rolebindings.go`
- [ ] `charts/kor/templates/role.yaml`
- [ ] `README.md`

nati-elmaliach:
This is a partial PR as I ran into challenges validating whether User or Group subjects exist. As you likely know, Kubernetes doesn’t store user or group information in its resources. I could use your input—am I missing something? How can we reliably verify the existence of a user or group?

Currently, a RoleBinding is considered unused if it references a non-existent Role, ClusterRole, or if none of its ServiceAccount subjects are valid.

We might consider merging this PR as is and opening a separate issue to address User and Group validation—I'll leave that decision to you.
## PR Checklist

- [x] This PR adds new code
- [ ] This PR adds docs

## GitHub Issue

Closes #334

## Notes for your reviewers
rolebinding-feature


pkg/kor/rolebindings.go
```diff
@@ -0,0 +1,164 @@
...
+
+	"github.com/yonahd/kor/pkg/common"
+	"github.com/yonahd/kor/pkg/filters"
+)
+
+//go:embed exceptions/rolebindings/rolebindings.json
+var roleBindingsConfig []byte
+
+// Convert a slice of names into a map for fast lookup
+func convertNamesToPresenseMap(names []string, _ []string, err error) (map[string]bool, error) {
```
yonahd, line 23: Maybe this can move into kor.go

pkg/kor/rolebindings.go
```diff
@@ -0,0 +1,164 @@
...
+
+		if rb.RoleRef.Kind == "Role" && !roleNames[rb.RoleRef.Name] {
+			unusedRoleBindingNames = append(unusedRoleBindingNames, ResourceInfo{Name: rb.Name, Reason: "RoleBinding references a non-existing Role"})
+			continue
+		}
+
+		if rb.RoleRef.Kind == "ClusterRole" && !clusterRoleNames[rb.RoleRef.Name] {
+			unusedRoleBindingNames = append(unusedRoleBindingNames, ResourceInfo{Name: rb.Name, Reason: "RoleBinding references a non-existing ClusterRole"})
+			continue
+		}
```
yonahd, line 104: nit; separate this into a function checkRoleReferences

pkg/kor/rolebindings.go
```diff
@@ -0,0 +1,164 @@
...
+
+		// If other kinds (Users/Groups) are used, we assume they exists for now
+		if len(serviceAccountSubjects) != len(rb.Subjects) {
+			continue
+		}
+
+		// Check if RoleBinding uses a valid service account
+		if !isUsingValidServiceAccount(serviceAccountSubjects, serviceAccountNames) {
+			unusedRoleBindingNames = append(unusedRoleBindingNames, ResourceInfo{Name: rb.Name, Reason: "RoleBinding references a non-existing ServiceAccount"})
+		}
```
yonahd, line 116: nit: separate into a function

yonahd (commented):
Overall looks good.
Regarding the users and groups I agree with your implementation, we will now mark as used as we prefer this option over false positives

doronkg:
One comment from a quick glance, relating both cmd & multi - 
although `rb` is quite intuitive, it's not an official kubectl shortname for RoleBindings.

For easier maintenance I'd suggest to remove it, see #300 for further reference.

doronkg:
As for users/groups - you're right, Kubernetes vanilla on its own doesn't store this data, as other distros like OpenShift do. 

We'll address that separately as you suggested, also relating for the future PR to map unused ClusterRoleBindings.

pkg/kor/rolebindings.go
nati-elmaliach: Actually i would like to keep it here for now, at least until we incorporate the user/group validation.

yonahd (approved):

yonahd:
LGTM
Thanks for the contribution!

PR #366, opened 2024-10-20
refactor: PDB label selection

### LINKED ISSUE #364: pdb not linked to deployment check giving false-positives
I have a PodDisruptionBudget using the following selector:
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: app-name
spec:
  maxUnavailable: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: app-name
      app.kubernetes.io/instance: app-instance
      
      
 I also have a deployment with the same selector in the helm manifest:
 apiVersion: apps/v1
kind: Deployment
metadata:
  name: app-name
spec:
  replicas: 2
  selector:
    matchLabels:
      app.kubernetes.io/name: app-name
      app.kubernetes.io/instance: app-instance
      
 
 
 It seems that the tool reports a false-positive that this pdb is not linked to any deployments/statefulsets

doronkg:
This PR changes the behavior of how `kor` maps unused PDBs.
Currently, it validates the PDB label selector against Deployments/StatefulSets direct metadata labels, while it should validate the template labels, as PDBs select pods and through `.metadata.ownerReferences` field finds the respective workload controllers and not the other way around.

For example, the following Deployment is valid, but `kor` will map it as unused, as `.metadata.labels` field is not matched, even if the template labels match:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app-name
spec:
  replicas: 2
  selector:
    matchLabels:
      app.kubernetes.io/name: app-name
      app.kubernetes.io/instance: app-instance
  template:
    metadata:
      labels:
        app.kubernetes.io/name: app-name
        app.kubernetes.io/instance: app-instance
```

In addition, added support for empty selectors & arbitrary pods.
- If a PDB has an empty selector, it's now considered as used (matches every pod)
> The behavior for an empty selector differs between the policy/v1beta1 and policy/v1 APIs for PodDisruptionBudgets. For policy/v1beta1 an empty selector matches zero pods, while for policy/v1 an empty selector matches every pod in the namespace.
- If a PDB has an empty selector, but no running pods are found in the namespace, it'll be considered as unused.
- PDBs support arbitrary pods (with no workload controllers), added validation for pods in addition to the Deployment/StatefulSet templates.
> You can use a PDB with pods controlled by another resource, by an "operator", or bare pods

## PR Checklist

- [x] This PR adds new code
- [ ] This PR adds docs

## GitHub Issue

Closes #364

## Notes for your reviewers

```
$ kor pdb --show-reason
Unused resources in namespace: "empty"
+---+---------------+----------------+--------------------------------------------------------------+
| # | RESOURCE TYPE | RESOURCE NAME  |                            REASON                            |
+---+---------------+----------------+--------------------------------------------------------------+
| 1 | Pdb           | app-name-empty | Pdb matches every pod (empty selector) but 0 pods run        |
| 2 | Pdb           | app-name2      | Pdb is not referencing any deployments, statefulsets or pods |
+---+---------------+----------------+--------------------------------------------------------------+
```

Based the feature on the official [PDB docs](https://kubernetes.io/docs/tasks/run-application/configure-pdb/).


pkg/kor/pdbs.go
```diff
@@ -52,34 +53,106 @@ func processNamespacePdbs(clientset kubernetes.Interface, namespace string, filt
...
+	deployments, err := clientset.AppsV1().Deployments(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return false, err
+	}
+
+	for _, deployment := range deployments.Items {
+		deploymentLabels := labels.Set(deployment.Spec.Template.Labels)
+		if labelSelector.Matches(deploymentLabels) {
+			return true, nil
 		}
```
yonahd, line 126:
```suggestion
	deploymentsWithLabel, err := clientset.AppsV1().Deployments(namespace).List(context.TODO(), metav1.ListOptions{
			LabelSelector: metav1.FormatLabelSelector(selector),
		})
		if len(deploymentsWithLabel.Items) > 0 {
			return true, nil
		}
```
Same for STS
This will cover 90% of the cases
The rest will be covered by validateMatchingWorkloads

yonahd (commented):
Looks good
One suggestion to speed it up

pkg/kor/pdbs.go
doronkg: This change wouldn't cover the following case:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app-name
spec:
  replicas: 0
  selector:
    matchLabels:
      app.kubernetes.io/name: app-name
      app.kubernetes.io/instance: app-instance
  template:
    metadata:
      labels:
        app.kubernetes.io/name: app-name
        app.kubernetes.io/instance: app-instance
```

yonahd: Won't this be covered by validate matching workloads?

yonahd (approved):

Issue #378, opened 2024-11-21
Bug: Duplicate printing of non existing namespaces


yonahd:
**Describe the bug**
Duplicate printing of non existing namespaces

**To Reproduce**
`kor -n fake1,fake2 configmap
namespace [fake1] not found
namespace [fake2] not found
namespace [fake1] not found
namespace [fake2] not found
kor version: vdev`

**Expected behavior**
kor -n fake1,fake2 configmap
namespace [fake1] not found
namespace [fake2] not found
kor version: vdev

**OS version, architecture and kor version**
Darwin x86_64
Kor version: 0.5.6

PR #383, opened 2024-11-28
Fix:	Removes duplicate printing of the included namespaces

almilaneze:

- Removes duplicate printing of the non existing included namespaces (which comes from the duplicated namespaces).
- Removes unnecessary variables initialization.

## PR Checklist

- [x] This PR adds new code
- [ ] This PR adds docs

## GitHub Issue
#378 
Closes [XX-XX]

### DIFF
--- a/pkg/filters/options.go
+++ b/pkg/filters/options.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -120,6 +121,8 @@ func (o *Options) Namespaces(clientset kubernetes.Interface) []string {
 		excludeNamespaces := o.ExcludeNamespaces
 
 		if len(o.IncludeNamespaces) > 0 {
+			slices.Sort(includeNamespaces)
+			includeNamespaces = slices.Compact(includeNamespaces)
 
 			for _, ns := range includeNamespaces {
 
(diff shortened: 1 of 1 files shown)


pkg/filters/options.go
```diff
@@ -120,6 +121,8 @@ func (o *Options) Namespaces(clientset kubernetes.Interface) []string {
 		excludeNamespaces := o.ExcludeNamespaces
 
 		if len(o.IncludeNamespaces) > 0 {
+			slices.Sort(includeNamespaces)
```
yonahd, line 124: Any idea why its coming duplicate?

pkg/filters/options.go
```diff
@@ -137,10 +140,6 @@ func (o *Options) Namespaces(clientset kubernetes.Interface) []string {
 				return
 			}
 
-			for _, ns := range namespaceList.Items {
```
yonahd, line 140: What is this removed for?

yonahd (commented):
@almilaneze looks like it fixes the issue
Couple of comments

pkg/filters/options.go
almilaneze: The removed code (_for loop_)  seems as not needed as it initializes with 'false' all the elements (available namespaces) of namespacesMap and right after that there is a analogical _for loop_ which changes the values to true. The logic of that section of the code is that initialization with 'true' values is the correct one. Both these were added in different commits (PRs) so I guess it was overlooked or there was an idea which was never implemented.

pkg/filters/options.go
almilaneze: Duplication is coming from pflag package and he way it parses argument of type string slice. Duplication is happening here:
https://github.com/spf13/pflag/blob/d5e0c0615acee7028e1e2740a11102313be88de1/string_slice.go#L50
I can't say for sure if it is a bug in the package or there is other logic behind it but looking at the definition of the filed 'changed' 
https://github.com/spf13/pflag/blob/d5e0c0615acee7028e1e2740a11102313be88de1/flag.go#L177
to me it more seems like an error in the logic.

yonahd (approved):

yonahd:
LGTM!
Merged

Issue #385, opened 2024-12-02
Wrongly marks a configmap as unused


almilaneze:
**Describe the bug**
When a configmap is used to provide environment variables to a initContainer (envFrom) this is not evaluated. As a result such a cm is marked as not used and if it is removed then on a new pod creation (for whatever reasons) it will fail as the corresponding configmap for providing environment vars to the  initContainer will be missing. 

**To Reproduce**

- have a pod with initContainer that gets env vars from a configmap
- check for unused configmaps in the namespace where the above pod is running

**Expected behavior**
Additional validation for initContainer's 'envFrom' section will be needed to check if it contains 'ConfigMapRef' section.

**Screenshots**

**OS version, architecture and kor version**
x86_64 GNU/Linux
Kor version: 0.1.14

**Additional context**

yonahd:
> Hi Yonah, I can take care of fixing this one.

Go for it.
Please make this fix also for the secrets and add a test for this case

PR #388, opened 2024-12-05
Check initContainers if they have env vars from a configmap or a secret

almilaneze:
Configmaps and secrets which were used only to provide env vars from them to an initContainer were wrongly evaluated as not used as there was not validation for this use case. Removal of such a cm or secret will prevent the POD from starting successfully.

## PR Checklist

- [x] This PR adds new code
- [ ] This PR adds docs

## GitHub Issue

Closes [#385 ]


pkg/kor/configmaps_test.go
```diff
@@ -95,6 +101,11 @@ func createTestConfigmaps(t *testing.T) *fake.Clientset {
 					ValueFrom: &corev1.EnvVarSource{ConfigMapKeyRef: &corev1.ConfigMapKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: configmap2.ObjectMeta.Name}}},
 				},
 			},
+			EnvFrom: []corev1.EnvFromSource{
```
yonahd, line 104: Nit: Can you make it used in a separate pod?

pkg/kor/secrets_test.go
```diff
@@ -78,6 +79,9 @@ func createTestSecrets(t *testing.T) *fake.Clientset {
 					ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: secret1.ObjectMeta.Name}}},
 				},
 			},
+			EnvFrom: []corev1.EnvFromSource{
```
yonahd, line 82: Same as above with cm

yonahd (commented):
Looks good overall. Minor comment

pkg/kor/configmaps_test.go
almilaneze: Done

pkg/kor/secrets_test.go
almilaneze: Done

yonahd (approved):
LGTM
Thanks!

PR #396, opened 2025-01-12
security: update some go modules

yonahd:


yonahd:
https://github.com/yonahd/kor/security/dependabot/6

PR #411, opened 2025-02-16
feat: Added --namespaced flag to filter namespaced/non-namespaced resources

doronkg:
Inspired by `kubectl api-resources --namespaced` flag, this PR adds the new flag to both `kor all` & `kor exporter` commands, allowing users to explicitly discover namespaced/non-namespaced resources.

Not using the flag == discovering both types of resources.
The change is also introduced in the Helm chart.

<details><summary>Examples</summary>

```
$ kor all -o yaml --group-by=resource
ConfigMap:
  default:
  - test
ServiceAccount:
  default:
  - test
ClusterRole:
  "":
  - test

$ kor all -o yaml --group-by=resource --namespaced
ConfigMap:
  default:
  - test
ServiceAccount:
  default:
  - test

$ kor all -o yaml --group-by=resource --namespaced=false
ClusterRole:
  "":
  - test
```
</details>

## PR Checklist

- [x] This PR adds new code
- [ ] This PR adds docs

## GitHub Issue

Closes [XX-XX]

### DIFF
--- a/charts/kor/Chart.yaml
+++ b/charts/kor/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: kor
 description: A Kubernetes Helm Chart to discover orphaned resources using kor
 type: application
-version: 0.1.18
-appVersion: "0.5.9"
+version: 0.1.19
+appVersion: "0.6.0"
 maintainers:
   - name: "yonahd"
     url: "https://github.com/yonahd/kor"
--- a/cmd/kor/all.go
+++ b/cmd/kor/all.go
@@ -17,8 +17,9 @@ var allCmd = &cobra.Command{
 		clientset := kor.GetKubeClient(kubeconfig)
 		apiExtClient := kor.GetAPIExtensionsClient(kubeconfig)
 		dynamicClient := kor.GetDynamicClient(kubeconfig)
+		scopeFlagUsed := cmd.Flags().Changed("namespaced")
 
-		if response, err := kor.GetUnusedAll(filterOptions, clientset, apiExtClient, dynamicClient, outputFormat, opts); err != nil {
+		if response, err := kor.GetUnusedAll(filterOptions, clientset, apiExtClient, dynamicClient, outputFormat, opts, scopeFlagUsed); err != nil {
 			fmt.Println(err)
 		} else {
 			utils.PrintLogo(outputFormat)
@@ -28,5 +29,6 @@ var allCmd = &cobra.Command{
 }
 
 func init() {
+	allCmd.PersistentFlags().BoolVar(&opts.Namespaced, "namespaced", true, "If false, non-namespaced resources will be returned, otherwise returning namespaced resources by default. If not used, both are returned")
 	rootCmd.AddCommand(allCmd)
 }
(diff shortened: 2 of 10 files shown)


pkg/kor/all.go
```diff
@@ -378,6 +380,13 @@ func GetUnusedAllNonNamespaced(filterOpts *filters.Options, clientset kubernetes
 }
 
 func GetUnusedAll(filterOpts *filters.Options, clientset kubernetes.Interface, apiExtClient apiextensionsclientset.Interface, dynamicClient dynamic.Interface, outputFormat string, opts common.Opts) (string, error) {
+	if NamespacedFlagUsed {
+		if opts.Namespaced {
+			return GetUnusedAllNamespaced(filterOpts, clientset, outputFormat, opts)
+		}
+		return GetUnusedAllNonNamespaced(filterOpts, clientset, apiExtClient, dynamicClient, outputFormat, opts)
```
yonahd, line 387: Shouldn't this return also namespaced resources?

charts/kor/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: kor
 description: A Kubernetes Helm Chart to discover orphaned resources using kor
 type: application
-version: 0.1.18
-appVersion: "0.5.9"
+version: 0.1.19
+appVersion: "0.6.0"
```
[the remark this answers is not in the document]
doronkg: If you refer to `appVersion`, I can lower it to `0.5.10` to include this change.
I thought of bumping to `0.6.0` where we could also merge #249 and release.

Let me know what to set.

cmd/kor/all.go
```diff
@@ -28,5 +29,6 @@ var allCmd = &cobra.Command{
 }
 
 func init() {
+	allCmd.Flags().BoolVar(&opts.Namespaced, "namespaced", true, "If false, non-namespaced resources will be returned, otherwise returning namespaced resources by default. If not used, both are returned")
```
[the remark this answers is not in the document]
doronkg: The default for the flag is `true` only if the flag is set (same as in `kubectl api-resources`). If it's not set, the command defaults to both. Take a look at the examples in the description.

pkg/kor/all.go
doronkg: No, the flow is as so:
1. Check is the flag is used
1.1. If it's used, check if it's true
1.1.1 If it's true, return namespaced resources
1.2. If it's not true, return non-namespaced resources
2. Return both (unless `--include-namespaces` is used)

yonahd: I think you set the default to true so it will return only namespaced resources if the flag is not used

yonahd: I'll try to make it clearer
The code assumes there are 3 different states
- flag is true
- flag is false
- flag does not exist

I think that because there is a default option 3 can never happen. I might be wrong but please check

doronkg: The code is aware to a scenario where the flag is not used, by passing the state through `cmd.Flags().Changed("namespaced")`, an example is shown in the description.

charts/kor/Chart.yaml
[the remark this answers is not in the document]
yonahd: Agree
After #249 we should bump

yonahd (approved):

PR #421, opened 2025-03-09
fix VolumeAttachment pluralisation in all.go

Yuni-sa:
Corrects the pluralization of `VolumeAttachment` in all.go.

## PR Checklist

- [x] This PR adds new code
- [ ] This PR adds docs

### DIFF
--- a/pkg/kor/all.go
+++ b/pkg/kor/all.go
@@ -363,13 +363,13 @@ func GetUnusedAllNonNamespaced(filterOpts *filters.Options, clientset kubernetes
 		resources[""]["Pv"] = getUnusedPvs(clientset, filterOpts).diff
 		resources[""]["ClusterRole"] = getUnusedClusterRoles(clientset, filterOpts).diff
 		resources[""]["StorageClass"] = getUnusedStorageClasses(clientset, filterOpts).diff
-		resources[""]["VolumeAttachments"] = getUnusedVolumeAttachments(clientset, filterOpts).diff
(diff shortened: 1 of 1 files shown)


Yuni-sa:
@doronkg

pkg/kor/all.go
```diff
@@ -369,7 +369,7 @@ func GetUnusedAllNonNamespaced(filterOpts *filters.Options, clientset kubernetes
 		appendResources(resources, "Pv", "", getUnusedPvs(clientset, filterOpts).diff)
 		appendResources(resources, "ClusterRole", "", getUnusedClusterRoles(clientset, filterOpts).diff)
 		appendResources(resources, "StorageClass", "", getUnusedStorageClasses(clientset, filterOpts).diff)
-		appendResources(resources, "VolumeAttachments", "", getUnusedVolumeAttachments(clientset, filterOpts).diff)
+		appendResources(resources, "VolumeAttachment", "", getUnusedVolumeAttachments(clientset, filterOpts).diff)
```
doronkg, line 372: Same for line 366 to align the output format

Yuni-sa: we should really open an issue to refactor and remove all of those hardcoded strings, they will get hard to maintain the more features we add and we might get more potential typos like those.

doronkg: Much agreed.

doronkg: We should also improve the "s" suffix as it doesn't much all plural forms.

doronkg (approved):

yonahd (approved):

PR #445, opened 2025-05-19
Feat: Get plural resource kinds from K8s

eden881:

This uses the K8s API to get plural resource kinds instead of relying on hard-coding an `s` after them.

For example, instead of printing `storageclasss`, which is wrong, now we print `storageclasses`.

## PR Checklist

- [x] This PR adds new code
- [ ] This PR adds docs

### DIFF
--- a/cmd/kor/root.go
+++ b/cmd/kor/root.go
@@ -46,16 +46,23 @@ var rootCmd = &cobra.Command{
 }
 
 var (
-	outputFormat  string
-	kubeconfig    string
-	opts          common.Opts
-	filterOptions = &filters.Options{}
+	outputFormat     string
+	kubeconfig       string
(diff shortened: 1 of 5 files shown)


yonahd:
Interesting. Was not familiar with this api
Can we maybe utilize it here?
https://github.com/yonahd/kor/blob/main/pkg%2Fkor%2Fmulti.go#L25

yonahd:
@eden881 do you plan on completing this one?

eden881:
@yonahd Should be ready to go :-)
I mocked `ResourceKindList` in the test case since a real Kubernetes client is never actually initialized, but it behaves the same.

eden881:
Rebased as requested @yonahd

PR #457, opened 2025-06-21
fix: skip kubeconfig initialization for utility and completion commands

chenrui333:

Update init process to skip kubeconfig initialization, which was added in https://github.com/yonahd/kor/pull/445

## PR Checklist

- [ ] This PR adds new code
- [ ] This PR adds docs

## Notes for your reviewers

seeing some completion generation issue when building 0.6.2 release, https://github.com/Homebrew/homebrew-core/pull/227766

### DIFF
--- a/cmd/kor/root.go
+++ b/cmd/kor/root.go
@@ -30,6 +30,14 @@ var rootCmd = &cobra.Command{
 	Long: `kor is a CLI to discover unused Kubernetes resources
 	kor can currently discover unused configmaps and secrets`,
 	Args: cobra.MinimumNArgs(1),
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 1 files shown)


eden881:
Just out of curiosity, why do you want to skip initialization for those specific commands?
I understand that they don't strictly need a client, but is it causing a specific issue for you?

jaudiger:
We got a similar problem on our side, with the upgrade to 0.6.2, we cannot anymore call `kor version` without having a KUBECONFIG: https://github.com/brioche-dev/brioche-packages/pull/712. Which impacts our CI for automatic check of new releases.

yonahd:
@chenrui333 I intend to create a release in the next few days
Would you like to complete this?

yonahd (approved):

PR #535, opened 2025-12-17
fix(helm): move imagePullPolicy and terminationMessagePath to proper …

yafanasiev:
…spec

`terminationMessagePath` and `imagePullPolicy` in Helm template for deployment are not indented correctly and end up in `toleration` instead of `container` spec.

## PR Checklist

- [x] This PR does something else

## Notes for your reviewers

Issue is quite easily reproduced with local `helm template` if `nodeSelector` is set in Helm values.


yafanasiev:
@yonahd done in 5ac0e3db59342144e31a4012b5bb40903f3e0e10
Sorry for skipping over the README, was in a bit of a rush

PR #563, opened 2026-01-28
feat: customizable RBAC templates

sleeyax:
## What this PR does / why we need it?

This pull request refactors how RBAC resources are generated in the Helm chart, making the configuration more flexible for consumers of the chart. 

See the linked issue for more context. For my use-case I can now install kor in our permission-limited cluster with the following config:

```yaml
rbac:
    create: 'role'
    rules:
      - apiGroups:
          - ''
        resources:
          - secrets
          - persistentvolumeclaims
          - pods
        verbs:
          - get
          - list
          - watch
          - delete
      - apiGroups:
          - networking.k8s.io
        resources:
          - ingresses
        verbs:
          - get
          - list
          - watch
```

```bash
$ helm upgrade --install kor-test helm --set kor.enabled=true --set "kor.cronJob.args={pvc\,secret,--delete,--no-interactive,--include-namespaces=development,--skip-namespace-validation}"
```

## PR Checklist

- [x] This PR modifies the helm chart

## GitHub Issue

Closes #561

## Notes for reviewers

Sorry for the messy commits. Took a bit of trial and error to land on a solution I'm satisfied with. Probably best to squash these changes.

PR #572, opened 2026-02-19
feat: add skip namespace validation option for RBAC filters

sleeyax:

Split from https://github.com/yonahd/kor/pull/563/.  

Currently, there's a hard-coded validation check in the code to verify that the supplied namespaces exist. However, clusters with limited permissions may not have the permission to list all namespaces. I added a flag to optionally disable this check for such cases.

## PR Checklist

- [x] This PR adds new code
- [ ] This PR adds docs

## GitHub Issue

Not applicable (I think). Feel free to create one retroactively.


sleeyax:
Any update on this? I apologize for opening this PR without creating an issue first. Would appreciate it if you could let me know if this is something that could potentially be merged or solved.

yonahd:
Hi @sleeyax
Been swamped with work recently.

We can potentially do it I'm a little concerned not to check at all.

Debating whether something like this is better

If you have get permissions on configmaps or serviceaccounts inside your target namespace, you can make a Get() call for a dummy name that you know does not exist (e.g., "exist-check-dummy").

The API server's response will reveal the truth:404 Not Found (Namespace does not exist): If the error message reads "namespaces 'target-ns' not found", the namespace is completely missing from the cluster.

404 Not Found (Namespace exists, resource doesn't): 
If the error message reads "configmaps 'exist-check-dummy' not found", the namespace definitely exists.

But this seems a bit fickle as well
Wdyt?

sleeyax:
Thanks for your response. That probably works too in my case. Could you provide a concrete example for me to check? 

> We can potentially do it I'm a little concerned not to check at all.

Just wanted to clarify that in my implementation in this PR it's an  **opt-in** option only. If you don't specify it, the namespaces are still checked.

PR #609, opened 2026-06-25
fix(clusterroles): don't ParseBool aggregation label values

### LINKED ISSUE #608: `retrieveUsedClusterRoles` is erroring, causing false-positive `ClusterRole`, `ClusterRoleBinding` and `RoleBinding` reports
This issue is well-described by https://github.com/yonahd/kor/pull/598, except that it doesn't only cause false-positives in `ClusterRole` reports, but also in `ClusterRoleBinding` and `RoleBinding` reports.

The PR is unmerged, and there doesn't seem to be a corresponding bug report. I'm raising this so that I can link to it.

RaphaelKorPro:
## Problem

`retrieveUsedClusterRoles` aborts the entire ClusterRole scan on any cluster running a controller that ships **aggregated ClusterRoles with non-boolean aggregation label values** — notably Emissary / Ambassador:

```
Failed to get clusterRoles: couldn't convert string to bool strconv.ParseBool: parsing "emissary-emissary-ingress-agent": invalid syntax
```

In `pkg/kor/clusterroles.go`, when a ClusterRole's label matches an aggregation selector, the code did:

```go
usedClusterRoles[clusterRole.Name], err = strconv.ParseBool(value)
if err != nil {
    return nil, fmt.Errorf("couldn't convert string to bool %v", err)
}
```

But `value` is the aggregation **label value** — an arbitrary string (e.g. `emissary-emissary-ingress-agent`), not a boolean. `ParseBool` errors and the whole non-namespaced ClusterRole scan bails out. The existing test only happened to use `"true"` as the value, so it never caught this.

## Fix

A ClusterRole whose label matches an aggregation selector is pulled into the aggregated (used) ClusterRole, so it is itself **used** — set it to `true` rather than parsing the value. `strconv` is no longer needed.

## Test

Adds `TestRetrieveUsedClusterRoles_NonBooleanAggregationLabel`, which reproduces the Emissary case (a non-boolean aggregation label value). It fails before the fix with the exact error above and passes after; existing clusterroles tests are unchanged.

### DIFF
--- a/pkg/kor/clusterroles.go
+++ b/pkg/kor/clusterroles.go
@@ -88,12 +88,7 @@ func retrieveUsedClusterRoles(clientset kubernetes.Interface, filterOpts *filter
 		for _, clusterRole := range clusterRoles.Items {
 			for label, value := range clusterRole.Labels {
 				if slices.Contains(aggregatedLabels, label+": "+value) {
-					// A ClusterRole whose label matches an aggregation selector is
-					// pulled into the aggregated (used) ClusterRole, so it is itself
-					// used. Previously this parsed the label *value* as a bool — but
-					// aggregation label values are arbitrary strings (e.g.
-					// "emissary-emissary-ingress-agent"), so ParseBool errored and
-					// aborted the entire ClusterRole scan.
+					// Aggregated into a used ClusterRole, so it is itself used.
 					usedClusterRoles[clusterRole.Name] = true
 					if clusterRole.AggregationRule == nil {
 						continue
--- a/pkg/kor/clusterroles.go
+++ b/pkg/kor/clusterroles.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"os"
 	"slices"
-	"strconv"
 
 	v1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


yonahd:
@RaphaelKorPro Can you fix this in 
ClusterRoleBinding and RoleBinding as well?

pkg/kor/clusterroles.go
```diff
@@ -89,10 +88,13 @@ func retrieveUsedClusterRoles(clientset kubernetes.Interface, filterOpts *filter
 		for _, clusterRole := range clusterRoles.Items {
 			for label, value := range clusterRole.Labels {
 				if slices.Contains(aggregatedLabels, label+": "+value) {
-					usedClusterRoles[clusterRole.Name], err = strconv.ParseBool(value)
-					if err != nil {
-						return nil, fmt.Errorf("couldn't convert string to bool %v", err)
-					}
+					// A ClusterRole whose label matches an aggregation selector is
```
yonahd: No need for this huge comment

RaphaelKorPro:
Done — trimmed the comment to a single line.

On ClusterRoleBinding/RoleBinding: I don't think there's an equivalent change to make here. This bug was specific to the **aggregation** path — `retrieveUsedClusterRoles` calls `strconv.ParseBool` on an aggregation label *value*. `AggregationRule` only exists on `ClusterRole` (Kubernetes only aggregates ClusterRoles, not Roles or any *Binding), and `clusterrolebindings.go` / `rolebindings.go` / `roles.go` have no `ParseBool` and no label-value parsing — they match by name / the `kor/used == "false"` string check, neither of which can hit this error.

Happy to add something if you had a specific case in mind — let me know.

yonahd (approved):

